PR #3, opened 2015-03-30
Rewrite in Go

samuelkarp:
This set of commits represents a complete rewrite in Go.

@euank, @aaithal: Care to take a look?


ecs-init/cache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+	defer publishedTarballReader.Close()
+
+	md5hash := md5.New()
+	tempFile, err := d.fs.TempFile("", "ecs-agent.tar")
+	if err != nil {
+		return err
+	}
+	defer tempFile.Close()
```
euank: You probably want to swap the order of your two defers so that you close then delete the file, not the reverse.

samuelkarp:
Oops!  This is a PR to the wrong branch.  Here's the right one: https://github.com/aws/amazon-ecs-init/pull/4

PR #5, opened 2015-04-01
Enable Agent updates and bump version to 1.0-1

samuelkarp:
@euank, @dangrd, @aaithal: Care to take a look?


euank:
LGTM. It would be nice if you could optionally use a fuzzy-selector for the agent version to tie to, but that doesn't have to be there now.

PR #6, opened 2015-04-15
Add missing gomock dependency to Godeps

samuelkarp:
Looks like I missed adding gomock as a vendored dependency (oops!).  Thanks @euank for showing me the right way to run godep (`godep save ./...`) such that it actually saves test dependencies as well.

r? @euank @aaithal


euank:
LGTM

PR #11, opened 2015-05-28
Update to support Agent 1.2.0 with Telemetry

samuelkarp:
I have not actually tested this yet but wanted to get the change out for review earlier rather than later.

r? @euank @aaithal


aaithal:
Looks good. I can test if execdriver mount breaks any of the current assumptions.

samuelkarp:
Thanks!  Let me know what works for you.  I'll also do my own testing today.

euank:
LGTM assuming the above mentioned testing doesn't uncover anything.

aaithal:
Tested that config changes work for docker 1.6.2. :ship:

PR #18, opened 2015-08-03
Update to support Agent 1.3.1

samuelkarp:
- Cache Agent version 1.3.1
- Read Docker endpoint from environment variable DOCKER_HOST if present

r? @euank @aaithal


aaithal:
lgtm

PR #22, opened 2015-09-03
Add default logging drivers

samuelkarp:
Add default logging drivers `json-file` and `syslog`.  Also includes a commit that better-handles user-supplied environment variables.

r? @euank


euank:
:+1:

PR #23, opened 2015-09-11
- Integrate ECS with openSUSE & SUSE Linux Enterprise Server

rjschwei:


samuelkarp:
This is great, thank you for contributing!  Can you let me know what the cgroup path and execdriver path are on openSUSE and SUSE Linux Enterprise Server?  These are currently [hard-coded to appropriate values for the Amazon Linux AMI](https://github.com/aws/amazon-ecs-init/blob/77ab8d443b8ba19c6f5fc8dd9e7b5b6525b501ff/ecs-init/config/common.go#L93-L99) and should be adjusted if appropriate.  I also have a few line-specific comments on the spec file.

packaging/suse/amazon-ecs-init.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
```
samuelkarp: I have a minor preference for the service being called `ecs` (as it is on the Amazon Linux AMI) or `amazon-ecs`.  I don't know that customers would expect to type `-init`.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via http://bugs.opensuse.org/
+#
+
+Name:           amazon-ecs-init
+Version:        1.3.1
```
samuelkarp: Are you packaging 1.3.1?  The current version we have is 1.4.0 and I'd like to make sure that whatever version indicated in the spec file matches the version of the Agent that gets loaded into Docker.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+Release:        0
+Summary:        Amazon EC2 Container Service Initialization
+License:        Apache-2.0
+Group:          System Environment/Base
+Url:            https://github.com/aws/amazon-ecs-init
+Source0:        %{name}-%{version}.tar.gz
+Source1:        %{name}.service
+BuildRequires:  go
+BuildRequires:  systemd
```
samuelkarp: Is this required for the `%service_*` macros?

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+Requires:       docker => 1.6.0
```
samuelkarp: Looks like this should be `>=`.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+Requires:       systemd
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+
+%description
+The Amazon Container Service initialization will start the ECS agent.
+The ECS agent runs in a container and is needed to support integration
+between the aws-cli ecs command line tool and an instance running in AWS EC2.
```
samuelkarp: Please change AWS EC2 to Amazon EC2.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+./scripts/gobuild.sh
+gzip -c scripts/amazon-ecs-init.1 > scripts/amazon-ecs-init.1.gz
```
samuelkarp: My colleague recommends using the SUSE post-build scripts to compress the man page rather than using gzip directly.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+install -d -m 755 %{buildroot}/%{_mandir}/man1
+install -d -m 755 %{buildroot}/%{_sbindir}
+install -d -m 755 %{buildroot}/%{_sysconfdir}/ecs
+install -m 644 scripts/amazon-ecs-init.1.gz %{buildroot}/%{_mandir}/man1
+install -m 755 amazon-ecs-init %{buildroot}/%{_sbindir}
+
+mkdir -p %{buildroot}/%{_unitdir}
+install -m 755 %SOURCE1 %{buildroot}/%{_unitdir}
+
+touch %{buildroot}/%{_sysconfdir}/ecs/ecs.config
```
samuelkarp: Can you add:

```
touch %{buildroot}/%{_sysconfdir}/ecs/ecs.config.json
```

This is an alternate way to configure the ECS Agent and both ways are supported by ecs-init.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+
```
samuelkarp: Can you add:

```
mkdir -p %{_localstatedir}/cache/ecs
touch %{_localstatedir}/cache/ecs/ecs-agent.tar
touch %{_localstatedir}/cache/ecs/state
```

This is part of how ecs-init manages loading the right Agent image into Docker.  When the RPM is installed or upgraded, we expect these files to exist but be empty.  After ecs-init runs the first time, it will populate both of these files.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+%files
+%defattr(-,root,root,-)
+%dir %{_sysconfdir}/ecs
+%doc CONTRIBUTING.md LICENSE NOTICE README.md
+%config(noreplace) %{_sysconfdir}/ecs/ecs.config
```
samuelkarp: Please add `%config(noreplace) %{_sysconfdir}/ecs/ecs.config.json`.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,78 @@
...
+%{_mandir}/man*/*
+%{_sbindir}/*
+%{_unitdir}/amazon-ecs-init.service
+
+%pre
+%service_add_pre amazon-ecs-init.service
```
samuelkarp: Do these macros handle restarting the service on upgrade when it's already enabled?  Also, does the service get enabled by default upon installing or does a user need to run `sudo systemctl enable ecs`?

samuelkarp:
I've just pushed #25 which should make it easier to configure cgroup and execdriver paths for openSUSE and SUSE Linux Enterprise Server.

samuelkarp:
@rjschwei, any update here?

packaging/suse/amazon-ecs-init.spec
rjschwei: Yes, systemd package provides the macros used later.

rjschwei:
I am working on incorporating all the suggestions, should be done today. Including the bump to 1.4.0-2

packaging/suse/amazon-ecs-init.spec
rjschwei: The macros handle the restart if the service is enabled. The service does not get enabled by default when the package gets installed. We enable the service as part of our image build process.

We will release an ECS image into the general catalog.

rjschwei:
OK, I think I got it all. Service has been renamed, spec file has been improved.

Can you please remove and merge if the changes are OK?

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,91 @@
...
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via http://bugs.opensuse.org/
+#
+
+%define short_name amazon-ecs
+Name:           amazon-ecs-init
+Version:        1.4.0.2
```
samuelkarp: Can you change this back to 1.4.0?  Our convention is to use the `Version` field to identify the version of the Agent and the `Release` field for changes to ecs-init itself or changes in the spec file.  In the case of `1.4.0-1` and `1.4.0-2`, the only difference is support for Docker 1.7.1 in the spec file.

samuelkarp:
@rjschwei Thanks!  My one outstanding concern here is with the cgroup and execdriver paths.  Can you verify what those are on openSUSE and SUSE Linux Enterprise Server?  Without those set correctly, the [CloudWatch Metrics](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/viewing_cloudwatch_metrics.html) support will not work.

packaging/suse/amazon-ecs-init.spec
rjschwei: If we do not encode the version somehow, and -x is not allowed in the version how would we differentiate 1.4.0-1 and 1.4.0-2 if the spec file remains at Version: 1.4.0

If I cannot make the differentiation between 1.4.0-1 and 1.4.0-2 in the spec file version declaration we will be stuck with always the first release. I don't think this is in our best interest. Adding the .2 at the end was the most direct way I could think of to capture the -X naming convention used for the tarball.

samuelkarp: I don't think you'd typically need to care about the difference from 1.4.0-1 to 1.4.0-2 as it's usually just spec file changes for the Amazon Linux AMI (e.g., changing the Requires for Docker when we have a new version to be released).  On the other hand, you have a good point.  If you'd prefer to leave it this way (and expect that it would usually end in .1 for you), then I will not consider this a blocker.

rjschwei:
Sorry not certain what you mean with "cgroup and execdriver paths" are these executables we are supposed to have?

packaging/suse/amazon-ecs-init.spec
rjschwei: If we do not need to care about -1 vs. -2 I am happy to drop the extra version :smile:

samuelkarp:
[cgroups](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt) are the underlying kernel features used to power Docker containers.  There is a cgroup hierarchy virtual filesystem available in userspace for browsing and manipulation of the cgroup hierarchy.  The location of this cgroup hierarchy virtual filesystem is not standardized; for example it is located at `/cgroup` on the Amazon Linux AMI but is located at `/sys/fs/cgroup` on Ubuntu.

The execdriver path is where Docker stores runtime information about the containers, including paths within the cgroup virtual filesystem that store the underlying cgroup configuration.  On the Amazon Linux AMI and on Ubuntu, this is located at `/var/run/docker/execdriver`.  I believe this path is hardcoded in Docker, so it should be the same unless you've patched Docker when building your RPM.

The ECS Agent accesses container runtime information from the execdriver path to map a given cotnainer to the associated cgroups, then accesses the cgroups to retrieve metrics for the running containers (CPU and memory utilization).  Both these directories must be mounted into the Agent container in order to make this feature work and performing the mounting is one of the responsibilities of ecs-init.  The paths are [defined](https://github.com/aws/amazon-ecs-init/blob/9600a0b540ec10b919897b0e345a0bf69089dca0/ecs-init/config/default_system_paths.go) for the Amazon Linux AMI already, and if the are different on openSUSE and SUSE Linux Enterprise Server we can control that with a build flag.

rjschwei:
cgroup information lives in /sys/fs/cgroup/

No modification to docker, thus /var/run/docker/execdriver on SUSE as well.

samuelkarp:
For now, if you want to keep using the [1.4.0-2 tarball](https://github.com/aws/amazon-ecs-init/archive/v1.4.0-2.tar.gz), your best bet is going to be patching [this line](https://github.com/aws/amazon-ecs-init/blob/92490594c3a7de7e8ce9444ced8190f35e1a3181/ecs-init/config/common.go#L94) to fix the path.  Going forward, we can add a build flag to change the path for SUSE which you can then use in your spec file without having to patch anything.  If you want to do this, I'd be happy to review it in this PR; otherwise I can do so and you can use it (though I probably won't be able to get to it this week).

rjschwei:
OK, for now I have added a SUSE specific patch to separate this commit from a change in the cgroups location.

packaging/suse/amazon-ecs-init.spec
```diff
@@ -0,0 +1,93 @@
...
+Patch1:         cgroup_location.patch
+BuildRequires:  go
+BuildRequires:  systemd
+Requires:       docker >= 1.6.0
+Requires:       systemd
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+ExclusiveArch:  %ix86 x86_64
+
+%description
+The Amazon Container Service initialization will start the ECS agent.
```
samuelkarp: Missed this before.  Can you change "Amazon Container Service" to "Amazon EC2 Container Service"?

samuelkarp:
Great!  I'd like to test this before merging.  Are [these](https://build.opensuse.org/package/show/Cloud:Tools/amazon-ecs-init) the correct RPMs?

packaging/suse/amazon-ecs-init.spec
rjschwei: On 09/27/2015 04:09 PM, Samuel Karp wrote:
>> +Summary:        Amazon EC2 Container Service Initialization
>> +License:        Apache-2.0
>> +Group:          System Environment/Base
>> +Url:            https://github.com/aws/amazon-ecs-init
>> +Source0:        %{name}-%{version}.tar.gz
>> +Source1:        %{short_name}.service
>> +Patch1:         cgroup_location.patch
>> +BuildRequires:  go
>> +BuildRequires:  systemd
>> +Requires:       docker >= 1.6.0
>> +Requires:       systemd
>> +BuildRoot:      %{_tmppath}/%{name}-%{version}-build
>> +ExclusiveArch:  %ix86 x86_64
>> +
>> +%description
>> +The Amazon Container Service initialization will start the ECS agent.
> 
> Missed this before.  Can you change "Amazon Container Service" to "Amazon EC2 Container Service"?
> 

Done


-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

rjschwei:
On 09/27/2015 04:10 PM, Samuel Karp wrote:
> Great!  I'd like to test this before merging.  Are [these](https://build.opensuse.org/package/show/Cloud:Tools/amazon-ecs-init) the correct RPMs?

Yes


-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

samuelkarp:
Finally got a chance to test this today and it looks great!  Thank you for the contribution!

euank:
:fireworks:

rjschwei:
On 10/02/2015 04:19 PM, Samuel Karp wrote:
> Finally got a chance to test this today and it looks great!  Thank you for the contribution!

Thanks for testing, thanks for the feedback and help to getting all the
pieces into place. Very happy to have this done, and just in time for
re:Invent.

I just published images to the general catalog in Amazon EC2. They can
be found with the name: suse-sles-12-v20151002-ecs-hvm-ssd-x86_64

samuelkarp:
I checked out the AMI and it looks to me like Docker and the ecs-init service do not start up automatically.  Is this intentional?

rjschwei:
On 10/02/2015 07:44 PM, Samuel Karp wrote:
> I checked out the AMI and it looks to me like Docker and the ecs-init
> service do not start up automatically. Is this intentional?

No, ecs-init is supposed to start automatically as is dockerd. I just
took a look at the image build setup and found the bug :( . While
amazon-ecs-init is set up sto start at boot time, dockerd was not. Thus
amazon-ecs-init start up will fail.

I fixed the image build setup. The good news is that we will have an
updated registration client in the next couple of weeks and thus we will
release new images. While this is a rather embarrassing and annoying bug
I am not certain it warrants new releases of the images. However, I can
probably be convinced otherwise.

-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

PR #25, opened 2015-09-16
Separate system paths for build logic control

samuelkarp:
This should help enable conditional compilation for other systems which have made different choices for the `cgroup` and `execdriver` paths.

This is intended to help with #23 if different `cgroup` and `execdriver` paths are needed there.

r? @euank


euank:
:+1:

PR #27, opened 2015-10-02
Do not overwrite /etc/init/ecs.conf on upgrades if it was modified

2opremio:
Our [Weave ECS AMIs](https://github.com/weaveworks/integrations/tree/master/aws/ecs) use a [modified version of `/etc/init/ecs.conf`](https://github.com/weaveworks/integrations/blob/master/aws/ecs/packer/to-upload/ecs.conf) to make sure that the ecs-agent talks to the weave proxy instead of docker directly.

However, we've noticed that `ecs.conf` gets overwritten on upgrades: weaveworks/integrations#4

This breaks the expected behaviour of our AMI, resulting in the `ecs-agent` talking to the docker daemon directly and bypassing our proxy.

### DIFF
--- a/packaging/amazon-linux-ami/ecs-init.spec
+++ b/packaging/amazon-linux-ami/ecs-init.spec
@@ -70,7 +70,7 @@ touch $RPM_BUILD_ROOT/%{cache_dir}/state
 
 %files
 %defattr(-,root,root,-)
-%{init_dir}/ecs.conf
+%config(noreplace) %{init_dir}/ecs.conf
 %{bin_dir}/amazon-ecs-init
 %{man_dir}/amazon-ecs-init.1.gz
 %config(noreplace) %ghost %{conf_dir}/ecs.config


samuelkarp:
Hi @2opremio, thanks for sending this.  We will not be accepting this change, as it removes our ability to modify the Upstart configuration during package upgrades.  However, you should be able to take advantage of an Upstart [override file](http://upstart.ubuntu.com/cookbook/#override-files) to get the desired behavior.

For example, if I create a file `/etc/init/ecs.override` that sets the `DOCKER_HOST` environment variable to a garbage value, I can observe ecs-init respecting the environment variable and correctly failing to work.

```
[ec2-user@ip-172-31-47-196 ~]$ cat /etc/init/ecs.override 
env DOCKER_HOST=unix:///dev/null
[ec2-user@ip-172-31-47-196 ~]$ cat /var/log/ecs/ecs-init.log.2015-10-02-21
2015-10-02T21:04:54Z [ERROR] dial unix /dev/null: connection refused
```

2opremio:
> it removes our ability to modify the Upstart configuration during package upgrades. 

Not really. If the file wasn't changed it should upgrade normally, see http://www-uxsup.csx.cam.ac.uk/~jw35/docs/rpm_config.html

Did you take this into account?

However, I didn't know about overrides, and they seem to be able to achieve what I need. If you still think that this change us not acceptable, I can try to use those.

Thanks!

samuelkarp:
Yes, I would recommend utilizing an override file.

PR #28, opened 2015-10-02
Changes to SUSE packaging for 1.5.0+

samuelkarp:
I've added a build flag for SUSE for builds after 1.5.0 and adjusted the SUSE patch for the 1.5.0 tarball.

@rjschwei, can you take a look?

### DIFF
--- a/ecs-init/config/suse_system_paths.go
+++ b/ecs-init/config/suse_system_paths.go
@@ -1,4 +1,4 @@
-// +build !suse
+// +build suse
 
 // Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
@@ -16,6 +16,6 @@
 package config
 
 const (
-	cgroupDirectory     = "/cgroup"
+	cgroupDirectory     = "/sys/fs/cgroup"
 	execDriverDirectory = "/var/run/docker/execdriver"
 )
--- a/packaging/suse/amazon-ecs-init.spec
+++ b/packaging/suse/amazon-ecs-init.spec
@@ -17,7 +17,7 @@
 
 %define short_name amazon-ecs
 Name:           amazon-ecs-init
-Version:        1.4.0
+Version:        1.5.0
 Release:        0
 Summary:        Amazon EC2 Container Service Initialization
 License:        Apache-2.0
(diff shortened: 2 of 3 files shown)


euank:
Is there a reason to have the patch file add `suse_system_paths.go` and also check it in? I'd prefer just checking it in, not the patch file.

samuelkarp:
Yes, that's so the existing [v1.5.0-1.tar.gz](https://github.com/aws/amazon-ecs-init/archive/v1.5.0-1.tar.gz) can be used rather than requiring the `dev` branch.

rjschwei:
I tend to agree with euank. If the release of 1.5.0 is only to accommodate the SUSE build we can certainly wait until the config change hist the master branch.

Additionally, while I understand that the build time configuration has a certain advantage, I am not certain that a runtime check of the path existing or not would not be sufficient, i.e. 

if os.Stat('/cgroups') {
  return '/cgroups
}
....

Or define an array with possible locations and have the if condition in a loop returning the one that is found. This approach is a small runtime hit, but avoids the maintenance of build config files. It's a trade off. I am OK with the build config switch as well. But we can certainly wait until the config file is in the master branch and then just drop the suse patch.

samuelkarp:
The 1.5.0 release has a number of things that would be great if SUSE could pick up (a new Agent version, not just the spec file you added).  I added the patch file (which is basically just that previous commit) to make it easier, but if you just want to cherry-pick that commit on top of the 1.5.0-1 tag, I can get rid of the patch file.  Please let me know what you think here.

With respect to checking for existence of `/cgroups/`: I think that's an interesting idea, but I worry about correctness here.  I had tried running one of your earlier RPMs on SUSE and Docker ended up creating that directory for me (Docker's default behavior when you specify a volume mount that doesn't exist on the host is to just make it for you).  I think I'd lean toward the build flag because we already know what to expect when building for a particular platform.

rjschwei:
Sorry for the slowness on my part re:Invent is consuming all my time.

If the addition of the build switch provides us with the more robust behavior, which is what it sounds like to me then we should go with that option.

So lets go wit this PR and then when the build switch is integrated into master we can drop the SUSE specific patch.

Thanks for looking after this

rjschwei:
Picked everything up, packages are building in OBS Cloud:Tools project

samuelkarp:
@rjschwei Thanks for taking a look!

PR #31, opened 2015-10-16
Initial deb packaging for Ubuntu Trusty

samuelkarp:
This is a very rough first cut at packaging for Ubuntu Trusty.
- ecs-init depends on Go 1.3 but 1.2 is the latest in the Trusty repo.  I had to grab a 1.3 build from a PPA.
- In the `control` file and upstart configuration, this depends on the `docker-engine` package produced by the Docker project.
- I have not tested what happens when Docker is upgraded, so it may not properly restart.
- lintian produces a few warnings for this package that I'm not quite sure how to fix.
- Building the deb requires signing.  Right now it's using a key with my identity (since I've put myself in the changelog), but this may not be the right approach.

r? @euank


ecs-init/config/default_system_paths.go
```diff
@@ -1,4 +1,5 @@
 // +build !suse
+// +build !ubuntu
```
euank: I think it's more idiomatic to put these on one line (`!suse,!ubuntu`), but I'm not certain. I'm fine with leaving it as is if you prefer this.

euank:
I'm not by any means an expert at debian packaging, but nothing stands out as wrong or scary here. If you can find someone more experienced here to look at it, that'd be awesome.
I do think we should figure out the signing stuff properly, but that's probably not be a blocker for merging this, just distribution.
I guess I'm a :ok_hand: on this, tending towards a :+1:. If you'd like me to look more closely and actually build it for myself, I could.

Also, it might be cool to have a target for building in Docker for the `debuild` bit.

NeMO84:
Any update here? 

This would be very helpful for helping create a custom AMI that integrates with ECS using Ubuntu.

samuelkarp:
@NeMO84 I'm glad you're interested.  I'm not an expert in deb packaging, so I'm looking for feedback if you have any.  I'd also like to resolve the issues mentioned in the description before I'm comfortable merging this.  However, you should be able to pull this yourself and build from it and I believe it should work.

hSATAC:
Looking forward to this +1

mittomen:
Hi, could you tell me if I can rely upon this change in the close future? It makes more sense to us to use trusty as a customized cluster node...

Since I already gone through all your changes I'll add my comments if you don't mind...

ecs.conf
```diff
@@ -0,0 +1,21 @@
+# Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mittomen: I think this file committed accidentally based on the makefile.

packaging/ubuntu-trusty/debian/dirs
```diff
@@ -0,0 +1,3 @@
+etc/ecs
+var/log/ecs
+var/lib/ecs
```
mittomen: Isn't that *var/lib/ecs/data*?

packaging/ubuntu-trusty/ecs.conf
```diff
@@ -0,0 +1,21 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+description "Amazon EC2 Container Service init"
+author "Amazon Web Services"
+start on (started docker and stopped cloud-final)
+
```
mittomen: I think adding *respawn* is a good idea to avoid throttling start in case of failure.

ecs.conf
samuelkarp: Oops!  Thanks for catching this!

packaging/ubuntu-trusty/debian/dirs
samuelkarp: `data` is currently the only thing we use in `var/lib/ecs` so that's probably reasonable to change.  I had just specified the whole of `var/lib/ecs` in case.

packaging/ubuntu-trusty/ecs.conf
samuelkarp: Could be.  So far we've left it out of the Upstart file that we use on the Amazon Linux AMI since ecs-init itself should be restarting the agent.  Getting the [respawn policy](http://upstart.ubuntu.com/cookbook/#respawn) right is a bit tricky and I want to make sure that we get it exactly right before I put it in.

samuelkarp:
@mittomen Thank you for the review comments, they're super helpful!  As I said to @NeMO84 before, I'd like to get the items outlined in the description addressed before I'm comfortable merging this.  Unfortunately I haven't had the time lately to come back to this and work on it, so I don't have an estimate of when it would be ready.

pilgrim2go:
Any update on this?
For some reason, our postgres 95 container failed in Amazon Linux. In Ubuntu, it works.
Is there any guide to setup  ECS instance with Ubutu?

Thanks.

samuelkarp:
@pilgrim2go I haven't had the chance to follow up more here.  Instructions for running the ECS agent on other Linux systems are in [our documentation](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html).

rickard-von-essen:
For anyone interested in testing this on Ubuntu 14.04 see my updated version of this which build `v1.13.0-1` and don't require signing. #70

rickard-von-essen:
This can be closed now since it was included in #72

samuelkarp:
@rickard-von-essen Yep, thanks!

PR #36, opened 2016-01-27
- Update SUSE packaging for 1.7.1

rjschwei:
+ remove cgroup_location.patch, path setting integrated into build
    with b51faa99b83004


samuelkarp:
This looks like the same diff as https://github.com/aws/amazon-ecs-init/pull/35.  I'm not sure why Github is even showing a diff, since `dev` already has these changes.

rjschwei:
Yes, I started the branch from master and not from dev. Obviously I only remembered 1/2 of what I reviewed the other day.

samuelkarp:
Thanks @rjschwei!

rjschwei:
OK, found the cause for my confusion; master already has ecs-init/config/default_system_paths.go and ecs-init/config/suse_system_paths.go to set up the SUSE specific paths, but still has the spec file for 1.5.0. As I was working on updating our package in OBS and then for SLE and looked at the released code from master (1.7.1-1) I completely missed the fact thet dev already had the changes to the spec file to catch up with the changes in master. Anyway, looks like I can proceed as planned, i.e. update the OBS package and SLE with 1.7.1-1 and get the changes for the spec file from dev, plus drop the patch.

Thanks

PR #37, opened 2016-01-28
Custom agent image name

jwerak:


samuelkarp:
Hi @veverjak, thanks for sending this.  Can you explain a bit about your use-case and why you'd want a configurable image?

veverjak:
Hi @samuelkarp sure. 

We have customized ecs-agent image (we needed to use custom docker log driver based on env var in task definition) and now we need to create custom build just to change image name. 

I believe there is no harm in having possibility to change ecs-agent docker image, but keeping the default one of course. 

what do you think?

samuelkarp:
Thanks.  Can you explain a bit more about what you've needed to do to customize the Agent rather than using the [`logConfiguration`](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_storage) in a task definition?

ap0:
We're using ECS to run a large number of containers via a homebrew job scheduler (think cron for ECS).  We need to be able to add a specific tag to the syslog configuration for every launch of a container (in this case, the ID of the task instance our scheduler's database).  We receive those in a syslog daemon we wrote and stream them into Kafka which makes its way back into our UI, and we can then break out the logs by each individual run of a job.  We did not want to create a new task definition for every single job we launch as we launch thousands of jobs a day and each one needs a unique tag; we are just using environment variable overrides in the `RunTask` API to instruct the modified agent to tag the logs appropriately for our use case.

This is what we're doing in the agent now: https://github.com/appuri/amazon-ecs-agent/commit/090a04b473f5e934a79d9a3253b19ab08ff2d01b#diff-e008975819a900a8494cf91881a8304a

Preferably we could specify log driver overrides like we can currently with environment variables, but this works for now until that capability is available.

samuelkarp:
Thanks, those details help a ton!  I'll circulate this internally and we'll figure out what we can do here.  In the meantime, are you having any difficulty building this locally so you can use it?

ap0:
I had to do some hackery (as you can see in the agent commit) to make it build, but I chalked that up to doing it in a Linux VM (normally a Mac user).  We have it built and running right now, though, and it works well.

veverjak:
Hi @samuelkarp, 

did you get any feedback from your team?

samuelkarp:
Apologies for leaving this for so long without a response.  I'm going to close this pull request today, but we're open to a contribution of this nature in the future.  Some specific things that we'd look for if someone is interested in contributing in this area in the future:

* Disable the agent update feature when using a custom image by setting `ECS_UPDATES_ENABLED` to `false` automatically
* Disable the cache population behavior (downloading the official image from S3) when using a custom image

PR #46, opened 2016-05-10
Remove unused execdriver and cgroup binds

samuelkarp:
These are unnecessary after https://github.com/aws/amazon-ecs-agent/pull/400.  I've built an RPM and tested it with an agent built from the tip of `dev`.

r? @aaithal @juanrhenals @richardpen


samuelkarp:
@rjschwei I've removed the SUSE-specific binds as well since those will not be necessary after we release the next version of the agent.

aaithal:
w00t! :shipit:

PR #47, opened 2016-05-24
Update support to agent 1.10.0

richardpen:
* Update to support Agent 1.10.0
* Add support to Docker 1.11
* Change default docker socket mount point to /var/run

r? @samuelkarp @aaithal @juanrhenals


samuelkarp:
:+1: It would have been nice for the Docker socket mount point change to be a separate commit, but this looks fine.

PR #49, opened 2016-06-29
Start agent with --net=host

aaithal:
`make test` passes. also verified that ECS Agent is started with `"NetworkMode"="Host"` on an instance with this change when I run `start ecs` and that I can ping the endpoints exposed by the agent.

r? @samuelkarp @juanrhenals @richardpen

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -24,14 +24,14 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


richardpen:
👍

samuelkarp:
:+1:

juanrhenals:
👍 LGTM

PR #50, opened 2016-07-01
cache: Download directly into cache directory

samuelkarp:
Fixes https://github.com/aws/amazon-ecs-agent/issues/420

Download the agent tarball directly into the cache directory such that there are no problems renaming the temporary download file into the final tarball name.

r? @aaithal @juanrhenals @richardpen


aaithal:
lgtm

richardpen:
👍

PR #66, opened 2016-09-27
Update to support agent version 1.13.0

nmeyerhans:


aaithal:
lgtm :shipit:

ecs-init/VERSION
```diff
@@ -0,0 +1 @@
+1.12.2
```
samuelkarp: Can you update this to `1.13.0` and run the `update-version.sh` script to generate a new version file?

nmeyerhans: done. please refresh this PR.

ecs-init/version/version.go
```diff
@@ -0,0 +1,27 @@
...
+package version
+
+// Version is the version of the ecs-init
+const Version = "1.13.0"
+
+// GitDirty indicates the cleanliness of the git repo when this ecs-init was built
+const GitDirty = true
+
+// GitShortHash is the short hash of this ecs-init build
+const GitShortHash = "c153a82"
```
samuelkarp: I think we missed copying this over in @richardpen's change set, but the `update-version.sh` script should set this to `UNKNOWN`.  Can you manually change it to `UNKNOWN` for now and then as a follow up create an issue to handle it?

richardpen:
👍

PR #70, opened 2016-11-11
deb packaging for Ubuntu Trusty - v1.13.0-1

rickard-von-essen:
This is a rebase of #31 onto `v1.13.0-1` and adding the following:

- Using package `golang-1.6`, no ppa required
- Option to skip signing .deb packages so anyone can build them.

build on Ubuntu 14.04 with:
```shell
apt-get install -y debhelper golang-1.6 git devscripts
DEB_SIGN=0 make clean ubuntu-trusty
```


rickard-von-essen:
Fixed one lintian warning and suppressed two which doesn't apply to this package.

IMHO this should be ready. It would be great to publish a deb package under releases, at least for next release.

I'll might add some build scripts to build all packages in docker.

rickard-von-essen:
Move this it #72 which targets the `dev` branch.

PR #72, opened 2016-11-17
deb packaging for Ubuntu Trusty - v1.13.1-1

rickard-von-essen:
This is a rebase of #31 onto `v1.13.1-1` and adding the following:

- Using package `golang-1.6`, no ppa required
- Option to skip signing .deb packages so anyone can build them.

build on Ubuntu 14.04 with:
```shell
apt-get install -y debhelper golang-1.6 git devscripts
DEB_SIGN=0 make clean ubuntu-trusty
```

Fixed one lintian warning and suppressed two which doesn't apply to this package.

IMHO this should be ready. It would be great to publish a deb package under releases, at least for next release.


rickard-von-essen:
Rebased on `v1.13.1-1`.

samuelkarp:
@rickard-von-essen Thank you for sending this, and my apologies for the very long delay.  Can you confirm that this contribution is licensed under the terms of the Apache 2.0 license?

@nmeyerhans Can you review this?

rickard-von-essen:
I hereby contribute my changes in this PR under the _Apache 2.0_ license.

packaging/ubuntu-trusty/debian/source/format
```diff
@@ -0,0 +1 @@
+3.0 (quilt)
```
nmeyerhans: Consider whether `3.0 (quilt)` is needed if we're not actually patching the source. Could this be `3.0 (native)` or event `1.0`?

scripts/gobuild.sh
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 set -x
 set -e
+export PATH=/usr/lib/go-1.6/bin:$PATH
```
nmeyerhans: Is this something we want to hardcode like this? Especially in a non distro version specific script, this doesn't seem ideal.

ecs.conf
```diff
@@ -0,0 +1,21 @@
+# Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
nmeyerhans: I suspect that this file was accidentally reintroduced, probably by the original PR by @samuelkarp. In any case, there shouldn't be a top-level `ecs.conf` in the repo.

packaging/ubuntu-trusty/debian/rules
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+
+# Uncomment this to turn on verbose mode.
+export DH_VERBOSE=1
+
+%:
+	dh $@
+
```
nmeyerhans: The `debian/rules` file also needs to take care of removing the `amazon-ecs-init` binary in the `clean` rule. Something like:

   clean:
      dh $@
      rm -f amazon-ecs-init

should work. We should probably also call `go clean`, possibly by way of our `gobuild.sh` script.

packaging/ubuntu-trusty/debian/source/format
rickard-von-essen: Native is only for programs specifically developed for Debian:

> After squeeze, it should say either:
>  - 3.0 (native) for native Debian packages or
>  - 3.0 (quilt) for everything else.

See https://www.debian.org/doc/manuals/maint-guide/dother.en.html#sourcef

Makefile
```diff
@@ -51,6 +54,17 @@ srpm: sources
 rpm: sources
 	rpmbuild -bb ecs-init.spec
 
+ubuntu-trusty:
+	cp packaging/ubuntu-trusty/ecs.conf ecs.conf
```
rickard-von-essen: Fixed!

(`/ecs.config` was probably added by mistake in the original code)

scripts/gobuild.sh
rickard-von-essen: Moved to `debian/rules`

ecs.conf
rickard-von-essen: Deleted

rickard-von-essen: And added to `.gitignore`

packaging/ubuntu-trusty/debian/rules
rickard-von-essen: I added a clean rule.

But not `go clean`, I don't see why that would be needed.

rickard-von-essen:
@nmeyerhans @samuelkarp This is ready for review. I have addressed the comment. It would be great to get this in.

rickard-von-essen:
@nmeyerhans ping..

nmeyerhans:
@rickard-von-essen Please rebase against the current `dev` branch and update the version in your `debian/changelog` file. Then we should be good to merge this.

rickard-von-essen:
Rebased onto `dev` and updated `debian/changelog`.

nmeyerhans:
Merging. Thank you for your contribution!

PR #73, opened 2016-11-17
Start the ECS Agent with --userns host

rickard-von-essen:
If you start your docker daemon with `--userns-remap=<val>` docker will map users inside containers to other uids on the host. Most notably this makes uid 0 inside a container != uid 0 on the host.

ECS agent needs access to the host network but you can't set `--network host` when you have `userns-remap`. By setting `--userns host` the ECS agent is started without userns remapping. This flag was added in Docker 1.11.

This required a update of the `go-dockerclient`.


rickard-von-essen:
Rebased...

Some feedback would be appreciated...

samuelkarp:
@rickard-von-essen Apologies for the extremely delayed response.

Can you change `go-dockerclient` to be the most recent version instead of the first version with support for `UsernsMode`?  That version vendors dependencies with import path rewriting, which we'd like to avoid if possible.

rickard-von-essen:
@samuelkarp no problem, I'll update this tomorrow.

rickard-von-essen:
The most recent `go-dockerclient` requires go 1.7 because of the use of `context`. Should I still go ahead with that update?

samuelkarp:
@rickard-von-essen Thanks.  We need to keep ecs-init on Go 1.6 for now, so we'll need to still use `golang.org/x/net/context` instead of `context`.  Can you pick the most recent version that still uses `golang.org/x/net/context`?

rickard-von-essen:
> Can you pick the most recent version that still uses `golang.org/x/net/context`

Yes, brb

rickard-von-essen:
Last version that doesn't depend on `context` seems to be fsouza/go-dockerclient@c342fd4c3d69bdc9767083cc61d11756e7323b64. I'll update to that.

ecs-init/docker/docker.go
```diff
@@ -94,7 +96,7 @@ func (c *Client) IsAgentImageLoaded() (bool, error) {
 
 // LoadImage loads an io.Reader into Docker
 func (c *Client) LoadImage(image io.Reader) error {
-	return c.docker.LoadImage(godocker.LoadImageOptions{image})
+	return c.docker.LoadImage(godocker.LoadImageOptions{image, context.Background()})
```
rickard-von-essen: I put `context.Background()` here or do you want `context.TODO()` instead?

samuelkarp: `context.TODO()` seems more appropriate here.

rickard-von-essen: Fixed!

ecs-init/docker/docker_test.go
```diff
@@ -94,7 +96,7 @@ func TestLoadImage(t *testing.T) {
 
 	mockDocker := NewMockdockerclient(mockCtrl)
 
-	mockDocker.EXPECT().LoadImage(godocker.LoadImageOptions{nil})
+	mockDocker.EXPECT().LoadImage(godocker.LoadImageOptions{nil, context.Background()})
```
samuelkarp: Looks like this should expect either `context.TODO()` or `gomock.Any()` instead of `context.Background()`.

rickard-von-essen: Fixed

samuelkarp:
Code looks good, we just need to test this now.

corbyshaner:
When is this scheduled to be released as an RPM?  It looks like docker-1.12.6-1.17.amzn1.x86_64 and ecs-init-1.14.0-2.amzn1.x86_64 are the latest.

samuelkarp:
Just to give an update here: I tested this with Docker 1.12.6 on Amazon Linux and saw that it was working correctly (inspecting showed me `"UsernsMode": "host"`).  We still need to test this against older versions of Docker (back to 1.6.0, since that's the oldest supported in our spec file right now).

nmeyerhans:
Merged. Thank you @rickard-von-essen!

PR #77, opened 2017-01-17
Update to support Agent 1.14.0

petderek:
Cache Agent 1.14.0


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 Name:           ecs-init
-Version:        1.13.1
+Version:        1.14.0
 Release:        2%{?dist}
```
nmeyerhans: Release should get reset to '1', since this is the first package release for 1.14.0.

PR #78, opened 2017-01-23
Add back-off/retry logic to docker ping.

aaithal:
We have a dependency on docker socket being ready to listen to requests for
any of the ecs-init actions to complete. However, on amazon linux, the docker
init script doesn't wait for the socket file to be ready and assumes that the
docker daemon is running when the pid file is created. This results in
ecs-init's pre-stat and start failing as there's a race for the docker socket
file to be ready.

This commit adds a back-off/retry policy around the docker Ping API to remediate
this issue.

Testing Done:
- [X] `make rpm` succeeds
- [X] `make gotest` succeeds
- [X] Manual testing: Installed the RPM on an Instance and simulated docker daemon starting and socket file not being ready for `5s`. Verified that the ECS Agent was still started by ecs init after 5s. This used to fail in previous versions 
```
 cat /var/log/ecs/ecs-init.log.2017-01-23-18
2017-01-23T18:55:05Z [INFO] Network error connecting to docker, backing off for '1.14777941s', error: dial unix /var/run/docker.sock: no such file or directory
2017-01-23T18:55:06Z [INFO] Network error connecting to docker, backing off for '2.282153551s', error: dial unix /var/run/docker.sock: no such file or directory
2017-01-23T18:55:08Z [INFO] Network error connecting to docker, backing off for '4.466145821s', error: dial unix /var/run/docker.sock: no such file or directory
2017-01-23T18:55:13Z [INFO] pre-start
2017-01-23T18:55:13Z [INFO] start
2017-01-23T18:55:13Z [INFO] Container name: /ecs-agent
2017-01-23T18:55:13Z [INFO] Removing existing agent container ID: 05c1de4af71f2d2613124885af7ab27fea34cc203346922857d99aa9d579dc7b
2017-01-23T18:55:13Z [INFO] Starting Amazon EC2 Container Service Agent
```


ecs-init/backoff/backoff.go
```diff
@@ -0,0 +1,78 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package backoff
+
+import (
+	"math"
+	mathrand "math/rand"
```
samuelkarp: Why alias this import?

ecs-init/backoff/backoff.go
```diff
@@ -0,0 +1,78 @@
...
+		max:            max,
+		jitterMultiple: jitterMultiple,
+		multiple:       multiple,
+		maxRetries:     maxRetries,
+		count:          0,
+	}
+}
+
+func (rb *retryBackoff) Duration() time.Duration {
+	ret := rb.current
```
samuelkarp: I take it that this is not intended to be thread-safe?  Might be worth documenting that a `Backoff` can only be accessed from one goroutine at a time.

ecs-init/version/version.go
```diff
@@ -24,4 +24,4 @@ const Version = "1.13.1"
 const GitDirty = true
 
 // GitShortHash is the short hash of this ecs-init build
-const GitShortHash = "UNKNOWN"
+const GitShortHash = "bb0c0a3"
```
samuelkarp: This change should be reverted.  Did this happen from `make gogenerate`?

ecs-init/docker/dependencies.go
```diff
@@ -32,22 +36,43 @@ type dockerclient interface {
...
+	return godocker.NewVersionedClient(endpoint, apiVersionString)
+}
+
+func newDockerClient(dockerClientFactory dockerClientFactory, pingBackoff backoff.Backoff) (dockerclient, error) {
 	dockerUnixSocketSourcePath, fromEnv := config.DockerUnixSocket()
 	if !fromEnv {
 		dockerUnixSocketSourcePath = "/var/run/docker.sock"
 	}
-	client, err := godocker.NewVersionedClient(config.UnixSocketPrefix+dockerUnixSocketSourcePath, "1.15")
+	client, err := dockerClientFactory.NewVersionedClient(
```
samuelkarp: At some point (does not need to be now) we should rip out this code.  go-dockerclient has a `NewVersionedClientFromEnv` that handles this properly.

ecs-init/docker/dependencies.go
```diff
@@ -32,22 +36,43 @@ type dockerclient interface {
...
+		config.UnixSocketPrefix+dockerUnixSocketSourcePath, "1.15")
 	if err != nil {
 		return nil, err
 	}
-	err = client.Ping()
+	for {
+		err = client.Ping()
+		if err != nil && isNetworkError(err) && pingBackoff.ShouldRetry() {
```
samuelkarp: Do we care specifically that it's a network error?  Should we just retry all errors?

ecs-init/docker/docker.go
```diff
@@ -32,6 +34,21 @@ const (
...
+	maxBackoffDuration = 5 * time.Second
+	// backoffJitterMultiple specifies the backoff jitter multiplier
+	// coefficient when pinging the docker socket
+	backoffJitterMultiple = 0.2
+	// backoffMultiple specifies the backoff multiplier coefficient when
+	// pinging the docker socket
+	backoffMultiple = 2
+	// maxRetries specifies the maximum number of retries for ping to return
+	// a successful response from the docker socket
+	maxRetries = 4
```
samuelkarp: Is this enough?

aaithal: As from the comment in line 64, this adds up to `12-14s`, which I think should be enough. Do you think it should be more?

ecs-init/docker/dependencies.go
aaithal: I'm modifying this to retry on net error and when we don't get 200 from docker socket. I don't think we should always retry this.

ecs-init/backoff/backoff.go
aaithal: Addressed this and other comments in the new [commit](https://github.com/aws/amazon-ecs-init/pull/78/commits/8dc38d74875365902787c331e09a458271234ce6).

ecs-init/docker/docker.go
samuelkarp: I don't have any data on how long it should be.  That's why I'm wondering.  If you think 12s is long enough for retry I'm okay with that.

aaithal: Let's make it at least 15. Will change this to '5'

ecs-init/docker/dependencies.go
```diff
@@ -32,22 +37,48 @@ type dockerclient interface {
...
-	client, err := godocker.NewVersionedClient(config.UnixSocketPrefix+dockerUnixSocketSourcePath, "1.15")
+	client, err := dockerClientFactory.NewVersionedClient(
+		config.UnixSocketPrefix+dockerUnixSocketSourcePath, "1.15")
 	if err != nil {
 		return nil, err
 	}
-	err = client.Ping()
+	for {
+		err = client.Ping()
+		if err != nil {
```
samuelkarp: I think this would be clearer like this:

```go
if err == nil {
	break
}
shouldRetry := (isNetworkError(err) || isRetryablePingError(err)) && pingBackoff.ShouldRetry()
if !shouldRetry {
	break
}
backoffDuration := pingBackoff.Duration()
log.Infof( /*...*/)
time.Sleep(backoffDuration)
```

That gets rid of the multiple levels of indentation and I think reads more clearly.

aaithal: done

ecs-init/docker/dependencies.go
```diff
@@ -96,3 +129,17 @@ var standardFS = &_standardFS{}
...
 }
+
+func isNetworkError(err error) bool {
+	_, ok := err.(*net.OpError)
+	return ok
+}
+
+func isRetryablePingError(err error) bool {
+	godockerError, ok := err.(*godocker.Error)
+	if ok && godockerError.Status != http.StatusOK {
```
richardpen: If you are checking against [this line](https://github.com/fsouza/go-dockerclient/blob/master/client.go#L392), is there any chance the `godockerError.Status == http.StatusOK`? Since it has been checked [here](https://github.com/fsouza/go-dockerclient/blob/master/client.go#L392).

aaithal: It's more of a future-proofing thing where I don't want us to be impacted by changes in this code base.

PR #81, opened 2017-02-13
Added upstart respawn to restart failing a service

rickard-von-essen:
Restart 10 times with 15 seconds delay if the service failes.

### DIFF
--- a/packaging/amazon-linux-ami/ecs.conf
+++ b/packaging/amazon-linux-ami/ecs.conf
@@ -16,6 +16,9 @@ description "Amazon EC2 Container Service init"
 author "Amazon Web Services"
 start on stopped rc RUNLEVEL=[345]
 
+respawn
+respawn limit 10 15
+
 pre-start exec /usr/libexec/amazon-ecs-init pre-start
 exec /usr/libexec/amazon-ecs-init start
 pre-stop exec /usr/libexec/amazon-ecs-init pre-stop


nmeyerhans:
Hi @rickard-von-essen, thanks for sending this in. I've been considering adding something like this recently as well. I'll take a look at this in more depth and try to merge this in time for our next release.

aaithal:
I am not against adding `respawn` in general. But, I've noticed that it doesn't address either a failing `pre-start` or `post-stop`. It would also mean that ECS Agent that exits with a terminal error code would be restarted because of the `respawn` policy. So, we should take that into consideration as well (Depending on the use-case it could be a good or a very bad thing).

rickard-von-essen:
@aaithal since the ECS Agent have crashed in the past/might crash in the future adding a respawn that likely stops me from getting call in the middle of the night it's definitely good. If it in an esoteric case is bad then I think you're already in D-S. 

If it doesn't  address `pre-start` or `post-stop` then there is something left to do. 

To summarize: if my service crashes is bad, if it doesn't automatically recover it's very bad.

aaithal:
@rickard-von-essen ECS Init is responsible for restarting the ECS Agent when it crashes/dies with a non-terminal error code. [This code](https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/engine/engine.go#L142) should be taking care of that. Have you seen instances where that's not the case? We would be interested in looking at any logs that you can provide if you've seen such behavior before.

`respawn` in ECS Init would re-spawn the ECS Init service and might or might not restart ECS Agent depending on the failure scenario (For example, if something is wrong with the Docker socket file, of the ECS Agent's state file, re-spawning ECS Init is not going to address that issue). As I said earlier, adding `respawn` to ECS Init might be enough to deal with some of these cases and is a good idea. However, there are some edge cases that we should be thinking about when doing this. For example, if there are invalid attributes specified in the ECS config file, would it result in unnecessary calls to `register-container-instances`? Does this mean that ECS Agent is not failing fast enough etc.

rickard-von-essen-iz:
> Have you seen instances where that's not the case?

We had two ECS Agents crash the same day, see aws/amazon-ecs-agent#707. The `ecs` service didn't restart it. We where running ECS Agent v1.13.1 if I remember correctly.

sparrc:
closing for inactivity

Issue #74, opened 2016-12-19
sudo start and stop ecs hangs


tcco:
I have ssh'd into a docker optimized linux distribution and tried to run both sudo start ecs and sudo stop ecs. 

Both of which never finish processing. I was curious if there are any other necessary steps to be able to run these commands.

samuelkarp:
> I was curious if there are any other necessary steps to be able to run these commands.

If you're using the ECS-optimized AMI, there are no extra steps.

The only thing I can think of offhand that would cause a hanging behavior would be something wrong with the Docker daemon (both `start ecs` and `stop ecs` interact with Docker).  Are you able to run `docker ps` on the instance or does that hang as well?

benfortuna:
I am also getting this behaviour. ```docker ps|info``` also hangs.

Same behaviour after the following:

`
$ sudo service docker restart
/bin/sh: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8)
Stopping docker:                                           [  OK  ]
Starting docker:        .                                  [  OK  ]`

nrcxcia:
I am also seeing this behavior, it causes some instances to fail to join the ecs cluster

spovisenko:
I had the same problem. I've checked clusters EC2 target instance and found that **Agent Connected** is **false**. Via SSH I performed next command on my problem instance:
- Execute `sudo yum update -y ecs-init`
- Docker was hanging too, so I've checked docker's log: `cat var/log/docker`  and found many warning like this:

> time="2017-04-11T09:46:39.666741390Z" level=info msg="libcontainerd: new containerd process, pid: 2565" 
time="2017-04-11T09:46:39.673254362Z" level=fatal msg="open /var/run/docker/libcontainerd/containerd/7c42eac12e10353bb454e39f3c0474135fd43d968403305949fa94acb9c5f01e/state.json: no such file or directory" 

- clean dir `/var/run/docker/libcontainerd/containerd/`  exec: `rm -rf  /var/run/docker/libcontainerd/containerd/`
- `service docker stop`
- `service docker start`
- `service docker status` should output `docker (pid  xxxx) is running...`
- `sudo reboot`
- check your EC2 instance via console, **Agent Connected** should be **true** and task is runing (is desired count of tasks > 0)

Done

samuelkarp:
[...]

PR #84, opened 2017-03-07
update to support agent 1.14.1

adnxn:
* Cache Agent 1.14.1
* Add Ubuntu Trusty packaging here (#72)


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -149,6 +149,8 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Mon Mar 6 2017 Adnan Khan <adnkha@amazon.com> - 1.14.1-1
+- Cache Agent version 1.14.1
```
jhaynes: We should capture the addition of Ubuntu Trusty packaging here (#74)

PR #87, opened 2017-04-06
Region Based Agent Update

cabbruzzese:
Adds functionality to download agent from location based on region. Discovers region from EC2 Metadata and queries against a list of available URLs. Supports falling back to a default URL.

This adds the AWS SDK as a dependency. Due to an update in goimports, this also includes raising the go version to 1.5.


ecs-init/config/common.go
```diff
@@ -29,8 +31,18 @@ const (
 	AgentLogFile = "ecs-agent.log"
 
 	UnixSocketPrefix = "unix://"
+
+	// Default region name
```
aaithal: From https://blog.golang.org/godoc-documenting-go-code

> The convention is simple: to document a type, variable, constant, function, or even a package, write a regular comment directly preceding its declaration, with no intervening blank line. Godoc will then present that comment as text alongside the item it documents.   

`DefaultRegionName` is the ...

ecs-init/config/common.go
```diff
@@ -29,8 +31,18 @@ const (
...
+	DefaultRegionName = "default"
 )
 
+var S3BucketMap = map[string]string{
```
aaithal: Few things here.

1. `S3BucketMap` is global scoped, it doesn't need to be 
2. Please make this a prefix map so that it constructs `"default"` -> `"https://s3.amazonaws.com/amazon-ecs-agent/"` and move `ecs-agent-v1.14.1.tar` to be a const so that the map can be used to get the prefix and the latest tarball is only typed once some where
3. Please remove entries for `"us-east-1"` and `"us-west-2"` from this map.

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
+	val, exists := S3BucketMap[regionName]
+	if !exists {
+		return S3BucketMap[DefaultRegionName]
+	}
+
+	return val
+}
+
+// Retreive tarball URL from S3 Bucket list by searching for region name
+func FindTarballUrl () string {
```
aaithal: Why can't this be in `AgentRemoteTarball` so that you can get rid of this method? Right now `AgentRemoteTarball` is calling this directly and doing nothing else. Let's avoid that pattern.

ecs-init/config/common.go
```diff
@@ -99,3 +110,42 @@ func DockerUnixSocket() (string, bool) {
 	// eg: Docker daemon restart
 	return "/var/run", false
 }
+
+// Find AZ information from Metadata. If error return a blank result
+func EC2MetadataAZ () string {
```
samuelkarp: There's a convenience function in the SDK called [`Region`](https://github.com/aws/aws-sdk-go/blob/db3e1e27b1ace4fc57be9c5cf7cea0566bd12034/aws/ec2metadata/api.go#L114-L123) that does this for you.  Please just use `metadata.Region()` instead of duplicating the logic.

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
```
aaithal: 1. `EC2MetadataAZ` is global scoped. It doesn't need to be
2. The comment needs to start with the method name as per my previous comment

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
+	return azName
+}
+
+//convert AZ name to Region name or default if blank is returned
+func AZToRegionName (azName string) string {
+	if azName == "" {
+		return DefaultRegionName
+	}
+
+	return azName[0:len(azName)-1]
```
aaithal: You need to check if the string length is at least `1` before doing this.

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
+
+	if err != nil {
+		return ""
+	}
+
+	return azName
+}
+
+//convert AZ name to Region name or default if blank is returned
+func AZToRegionName (azName string) string {
```
aaithal: Same as above:

1. `AZToRegionName ` is global scoped. It doesn't need to be
2. The comment needs to start with the method name as per my previous comment

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
+	if azName == "" {
+		return DefaultRegionName
+	}
+
+	return azName[0:len(azName)-1]
+}
+
+// Get Bucket from list of S3 Buckets by region name or default if key is not found
+func GetS3BucketMapByRegion(regionName string) string {
```
aaithal: Same as above:

1. `GetS3BucketMapByRegion ` is global scoped. It doesn't need to be
2. The comment needs to start with the method name as per my previous comment

ecs-init/config/common.go
```diff
@@ -99,3 +111,42 @@ func DockerUnixSocket() (string, bool) {
...
+	val, exists := S3BucketMap[regionName]
```
aaithal: Convention is to use `ok` instead of `exists`. Makes it very clear that you're looking up a map

ecs-init/config/common.go
cabbruzzese: The intention was to add an extra abstraction to support additional configurations in the future. As this is not being addressed right now I'll update to remove the superfluous function.

samuelkarp:
@cabbruzzese Can you look at the failing tests in Travis?  Also, can you explain why you raised the minimum Go version to 1.6?

ecs-init/Godeps/Godeps.json
```diff
@@ -15,8 +126,78 @@
 			"Rev": "c9ad0ce23f68428421adfc6ced9e6123f54788a5"
 		},
 		{
+			"ImportPath": "github.com/fsouza/go-dockerclient/external/github.com/Sirupsen/logrus",
```
samuelkarp: This looks like an old version of go-dockerclient; the current version no longer has its dependencies vendored.  We should not use this version of go-dockerclient with the vendored, rewritten imports.

ecs-init/config/common.go
```diff
@@ -99,3 +117,28 @@ func DockerUnixSocket() (string, bool) {
...
+
+	if err != nil {
+		return DefaultRegionName
+	}
+
+	return regionName
+}
+
+// getS3BucketMapByRegion fetches the bucket URI from list of S3 Buckets by region name or default if key is not found
+func getS3BucketMapByRegion(regionName string) string {
```
samuelkarp: This should be named `getS3BucketForRegion` or (more generically) `getBaseLocationForRegion`.

ecs-init/config/common.go
```diff
@@ -99,3 +117,28 @@ func DockerUnixSocket() (string, bool) {
 	// eg: Docker daemon restart
 	return "/var/run", false
 }
+
+// ec2MetadataRegion finds the Region name from Metadata. If an error occurs fetching the region the default region name
+// is returned.
+func ec2MetadataRegion () string {
```
samuelkarp: Name this `getRegion`.  It's an implementation detail that it comes from EC2 instance metadata.

ecs-init/config/common.go
```diff
@@ -99,3 +117,28 @@ func DockerUnixSocket() (string, bool) {
...
+	// Find Region name from Metadata. If error return a blank result
+	sessionInstance := session.Must(session.NewSession())
+	metadata := ec2metadata.New(sessionInstance)
+	regionName, err := metadata.Region()
+
```
samuelkarp: Remove blank line

ecs-init/config/common.go
```diff
@@ -32,12 +32,13 @@ const (
 
 	UnixSocketPrefix = "unix://"
 
-	// Default region name
+	//DefaultRegionName is the name of the region to fall back to if no entry for the region name is found in the
```
aaithal: [multiple occurrences] Comments should have a space between `//` and the description: https://golang.org/doc/effective_go.html#commentary

`// DefaultRegionName`

aaithal:
@cabbruzzese Can you please fix the travis build as well?

samuelkarp:
@cabbruzzese It looks like it's still failing in Travis.  Can you fix those failures?

ecs-init/Godeps/Godeps.json
cabbruzzese: This was the same go-dockerclient version. There is a known bug with Godep v79 that vendors nested dependencies that have already been vendored. (https://github.com/tools/godep/issues/498)

Reverting to Godep v29 (latest version used to ecs-init) fixes this. Update pending.

Additionally, the latest versions of go-dockerclient do not support go 1.5. I would like to get both of these dependencies up to date and migrate to the vendor/ folder in a future change, but it is out of scope for this PR.

cabbruzzese:
Travis build issue resolved. Required refactor to treat fetching region as a dependency. Please review again

ecs-init/cache/cache.go
```diff
@@ -104,8 +123,10 @@ func (d *Downloader) DownloadAgent() error {
 	calculatedMd5SumString := fmt.Sprintf("%x", calculatedMd5Sum)
 	log.Debugf("Expected %s", publishedMd5Sum)
 	log.Debugf("Calculated %s", calculatedMd5SumString)
+	region := d.getRegion()
+	agentRemoteTarball := config.AgentRemoteTarball(region)
 	if publishedMd5Sum != calculatedMd5SumString {
-		err = fmt.Errorf("mismatched md5sum while downloading %s", config.AgentRemoteTarball())
+		err = fmt.Errorf("mismatched md5sum while downloading %s", agentRemoteTarball)
```
aaithal: you can combine these into a single line

Makefile
```diff
@@ -29,7 +29,7 @@ static:
 	./scripts/gobuild.sh
 
 gotest:
-	GOPATH="$(shell pwd)/ecs-init/Godeps/_workspace:$(GOPATH)" go test -short -v -cover ./...
+	GOPATH="$(shell pwd)/ecs-init/Godeps/_workspace:$(GOPATH)" go test -short -v -cover -timeout 10m ./...
```
aaithal: Can  you set this to something lower? I don't think it needs to be greater than `10s` even

ecs-init/cache/cache.go
```diff
@@ -61,6 +67,19 @@ func (d *Downloader) fileNotEmpty(filename string) bool {
 	return fileinfo.Size() > 0
 }
 
+// getRegion finds region from metadata and caches for the life of downloader
+func (d *Downloader) getRegion() string {
+	if d.region == "" {
+		region, ok := d.metadata.Region()
```
aaithal: This should be `region, err := `.. `ok` is used when a `bool` is being returned.

ecs-init/cache/cache.go
```diff
@@ -61,6 +67,19 @@ func (d *Downloader) fileNotEmpty(filename string) bool {
...
+		if ok != nil {
+			region = config.DefaultRegionName
+		}
+		d.region = region;
```
aaithal: Can you add a debug log line here for the region retrieved?

ecs-init/cache/cache_test.go
```diff
@@ -124,11 +125,14 @@ func TestDownloadAgentDownloadMD5Failure(t *testing.T) {
 	mockGetter := NewMockhttpGetter(mockCtrl)
 
 	mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700)
-	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5()).Return(nil, errors.New("test error"))
+	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5(config.DefaultRegionName)).Return(nil, errors.New("test error"))
```
aaithal: Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -145,12 +149,15 @@ func TestDownloadAgentReadPublishedMd5Failure(t *testing.T) {
 	mockFS := NewMockfileSystem(mockCtrl)
 	mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700)
 	mockGetter := NewMockhttpGetter(mockCtrl)
-	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5()).Return(md5response, nil)
+	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5(config.DefaultRegionName)).Return(md5response, nil)
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -166,13 +173,16 @@ func TestDownloadAgentDownloadTarballFailure(t *testing.T) {
...
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -191,13 +201,16 @@ func TestDownloadAgentDownloadTarballNon200(t *testing.T) {
...
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -217,14 +230,17 @@ func TestDownloadAgentTempFile(t *testing.T) {
...
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -244,9 +260,9 @@ func TestDownloadAgentCopyFailure(t *testing.T) {
...
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -256,10 +272,13 @@ func TestDownloadAgentCopyFailure(t *testing.T) {
 	mockFS.EXPECT().TeeReader(tarballResponse.Body, gomock.Any())
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -279,9 +298,9 @@ func TestDownloadAgentMD5Mismatch(t *testing.T) {
 	mockFS := NewMockfileSystem(mockCtrl)
 	mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700)
 	mockGetter := NewMockhttpGetter(mockCtrl)
-	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5()).Return(md5response, nil)
+	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5(config.DefaultRegionName)).Return(md5response, nil)
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -291,10 +310,13 @@ func TestDownloadAgentMD5Mismatch(t *testing.T) {
 	mockFS.EXPECT().TeeReader(tarballResponse.Body, gomock.Any())
 	mockFS.EXPECT().Copy(tempfile, gomock.Any()).Return(int64(0), nil)
 	mockFS.EXPECT().Remove(tempfile.Name())
+	mockMetadata := NewMockinstanceMetadata(mockCtrl)
+	mockMetadata.EXPECT().Region().Return(config.DefaultRegionName, nil)
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -315,9 +337,9 @@ func TestDownloadAgentSuccess(t *testing.T) {
 	mockFS := NewMockfileSystem(mockCtrl)
 	mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700)
 	mockGetter := NewMockhttpGetter(mockCtrl)
-	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5()).Return(md5response, nil)
+	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5(config.DefaultRegionName)).Return(md5response, nil)
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/cache/cache_test.go
```diff
@@ -332,10 +354,13 @@ func TestDownloadAgentSuccess(t *testing.T) {
 	})
 	mockFS.EXPECT().Copy(tempfile, gomock.Any()).Return(int64(0), nil)
 	mockFS.EXPECT().Rename(tempfile.Name(), config.AgentTarball())
+	mockMetadata := NewMockinstanceMetadata(mockCtrl)
+	mockMetadata.EXPECT().Region().Return(config.DefaultRegionName, nil)
```
aaithal: Same as above. Can we be more strict and the check the ordering here (Using [`gomock.InOrder`](https://godoc.org/github.com/golang/mock/gomock#InOrder))?

ecs-init/config/common.go
```diff
@@ -118,20 +115,6 @@ func DockerUnixSocket() (string, bool) {
...
-	if err != nil {
-		return DefaultRegionName
-	}
-
-	return regionName
-}
-
 // getBaseLocationForRegion fetches the bucket URI from list of S3 Buckets by region name or default if key is not found
 func getBaseLocationForRegion(regionName string) string {
 	val, ok := s3BucketMap[regionName]
```
aaithal: I'd prefer to rename `val` as `s3BucketURL`

ecs-init/config/common.go
```diff
@@ -32,15 +32,18 @@ const (
...
+	AgentFilename = "ecs-agent-v1.14.1.tar"
+
+	// DefaultRegionName is the name of the region to fall back to if no entry for the region name is found in the
+	// S3BucketMap.
 	DefaultRegionName = "default"
 )
 
-//s3BucketMap provides a mapping of region names to specific URI's for the region.
+// s3BucketMap provides a mapping of region names to specific URI's for the region.
 var s3BucketMap = map[string]string{
```
aaithal: Can you rename `s3BucketMap` to `regionToS3BucketURL`? It's good to avoid the type of the variable when naming it. Also, makes it more meaningful

cabbruzzese:
Updated based on feedback. Added InOrder expectation for cache tests

ecs-init/cache/cache.go
```diff
@@ -70,8 +70,9 @@ func (d *Downloader) fileNotEmpty(filename string) bool {
 // getRegion finds region from metadata and caches for the life of downloader
 func (d *Downloader) getRegion() string {
 	if d.region == "" {
-		region, ok := d.metadata.Region()
-		if ok != nil {
+		region, err := d.metadata.Region()
+		if err != nil {
+			log.Debug("Could not retrieve EC2 Metadata. %s", err)
```
aaithal: Can change the error message to be `"Could not retrieve the region from EC2 Instance Metadata"`? Also, this should be a warning, not debug.

cabbruzzese:
Updated log message

ecs-init/Godeps/Godeps.json
samuelkarp: go-dockerclient does not have a `vendor` directory.  This old version of go-dockerclient has an `external` directory that is not excluded because it's called `external` (and the imports are all rewritten within go-dockerclient to use that `external` package), so I don't think we're running into https://github.com/tools/godep/issues/498.

Is there not a newer version of go-dockerclient that (1) does not have a `vendor` (or an `external`) package and (2) works with Go 1.5?

samuelkarp: As per our offline sync: We should *not* merge this with the packages under `external`.  We can upgrade to Go 1.7, so if it's necessary to do that we should get that done first.

ecs-init/cache/cache.go
```diff
@@ -61,6 +67,20 @@ func (d *Downloader) fileNotEmpty(filename string) bool {
 	return fileinfo.Size() > 0
 }
 
+// getRegion finds region from metadata and caches for the life of downloader
+func (d *Downloader) getRegion() string {
+	if d.region == "" {
+		region, err := d.metadata.Region()
+		if err != nil {
+			log.Warn("Could not retrieve the region from EC2 Instance Metadata. %s", err)
```
aaithal: Also, if your format string is `%s`, please use `err.Error()` instead of `err`

ecs-init/cache/cache.go
```diff
@@ -25,24 +25,30 @@ import (
...
+	"github.com/aws/aws-sdk-go/aws/session"
 	log "github.com/cihub/seelog"
 )
 
 const (
 	orwPerm = 0700
 )
 
-// Downloader is resposible for cache operations relating to downloading the agent
 type Downloader struct {
```
samuelkarp: Why'd you remove the godoc comment?

ecs-init/cache/cache.go
```diff
@@ -25,24 +25,30 @@ import (
...
+	getter   httpGetter
+	fs       fileSystem
+	metadata instanceMetadata
+
+	region string
 }
 
 // NewDownloader returns a Downloader with default dependencies
 func NewDownloader() *Downloader {
+	sessionInstance := session.Must(session.NewSession())
```
samuelkarp: `session.Must` will panic if it can't create the session.  Use `session.NewSession` and check for an error.  If there is an error, we can fall-back to not querying `ec2metadata` and just use the default bucket.

ecs-init/cache/cache.go
```diff
@@ -61,6 +67,20 @@ func (d *Downloader) fileNotEmpty(filename string) bool {
 	return fileinfo.Size() > 0
 }
 
+// getRegion finds region from metadata and caches for the life of downloader
+func (d *Downloader) getRegion() string {
+	if d.region == "" {
```
samuelkarp: Reverse the conditional so we don't need this much indentation:

```go
if d.region != "" {
	return d.region
}
// do everything else
```

I don't think there's any concurrency going on here, but a note for other times you'll do this pattern: checking for something like this or mutating it normally will require a lock to prevent race conditions.

ecs-init/cache/cache_test.go
```diff
@@ -122,13 +123,18 @@ func TestDownloadAgentDownloadMD5Failure(t *testing.T) {
 
 	mockFS := NewMockfileSystem(mockCtrl)
 	mockGetter := NewMockhttpGetter(mockCtrl)
+	mockMetadata := NewMockinstanceMetadata(mockCtrl)
 
-	mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700)
-	mockGetter.EXPECT().Get(config.AgentRemoteTarballMD5()).Return(nil, errors.New("test error"))
+	gomock.InOrder(
+		mockFS.EXPECT().MkdirAll(config.CacheDirectory(), os.ModeDir|0700),
+		mockMetadata.EXPECT().Region().Return(config.DefaultRegionName, nil),
```
samuelkarp: (multiple occurrences) For the positive case of `Region()` returning a real value (as opposed to there being some error), please use a different return value than `config.DefaultRegionName`.  If you use `config.DefaultRegionName` it'll be more difficult to distinguish if the correct region is being used, since that's the fall-back when an error occurs.

Also: there needs to be a negative test case in here where `Region()` returns an error.  I don't see one.

ecs-init/cache/dependencies_mock_test.go
```diff
@@ -17,10 +17,9 @@
 package cache
 
 import (
-	io "io"
 	http "net/http"
 	os "os"
-
+	io "io"
```
samuelkarp: These look out of order.  Can you ensure that they're ordered by goimports?

ecs-init/config/common.go
```diff
@@ -29,8 +29,21 @@ const (
 	AgentLogFile = "ecs-agent.log"
 
 	UnixSocketPrefix = "unix://"
+
+	// AgentFileName is the filename, including version number, of the agent to be downloaded.
```
samuelkarp: Comment case does not match variable case.

ecs-init/Godeps/Godeps.json
cabbruzzese: This is being addressed in #94. Will rebase from that once merged.

ecs-init/cache/dependencies_mock_test.go
```diff
@@ -220,3 +219,35 @@ func (_m *MockfileSizeInfo) Size() int64 {
...
+func (_m *MockinstanceMetadata) Region() (string, error) {
+	ret := _m.ctrl.Call(_m, "Region")
+	ret0, _ := ret[0].(string)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+func (_mr *_MockinstanceMetadataRecorder) Region() *gomock.Call {
+	return _mr.mock.ctrl.RecordCall(_mr.mock, "Region")
+}
```
cabbruzzese: Appears to be there.

ecs-init/cache/cache_test.go
cabbruzzese: Unforunately with only two regions, the default case is the normal path. Currently this works as an exception list, only including custom values for special regions that override the default. As additional regions are added these tests should be added. For now, this ambiguity is the result of attempting to de-duplicate a previous change that included specifying another region as well as a default.

The negative test case has been added.

cabbruzzese:
Updated based on feedback

ecs-init/cache/cache.go
```diff
@@ -34,16 +36,29 @@ const (
...
+
+	// NewSession can fail. If metadata cannot be initialized the region string is populated with the default value
+	// to prevent future calls to retrieve the region from metadata
+	sessionInstance, err := session.NewSession()
+	if err != nil {
+		downloader.region = config.DefaultRegionName
+	} else {
+		downloader.metadata = ec2metadata.New(sessionInstance)
+	}
+	return
```
aaithal: Please delete this line, or move it to inside the `if err != nil {` block so that you can get rid of `else`

ecs-init/config/common_test.go
```diff
@@ -45,3 +45,20 @@ func TestDockerUnixSocketWithDockerHost(t *testing.T) {
 		t.Error("DockerUnixSocket() should read from envrionment variable DOCKER_HOST, when DOCKER_HOST is set")
 	}
 }
+
+func TestGetS3BucketMapByRegion(t *testing.T) {
+	s3Bucket := getBaseLocationForRegion(DefaultRegionName)
```
aaithal: With go 1.7, you can rewrite this as table driven tests. More info and examples:

* https://blog.golang.org/subtests
* https://github.com/golang/go/wiki/TableDrivenTests
* https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

ecs-init/cache/cache.go
```diff
@@ -55,9 +54,9 @@ func NewDownloader() (downloader *Downloader) {
 	sessionInstance, err := session.NewSession()
 	if err != nil {
 		downloader.region = config.DefaultRegionName
-	} else {
-		downloader.metadata = ec2metadata.New(sessionInstance)
+		return
 	}
+	downloader.metadata = ec2metadata.New(sessionInstance)
 	return
```
aaithal: This still needs to be deleted. We don't do blank returns in golang.

ecs-init/cache/cache.go
```diff
@@ -34,16 +36,28 @@ const (
...
 
 // NewDownloader returns a Downloader with default dependencies
 func NewDownloader() *Downloader {
-	return &Downloader{
+	downloader := &Downloader{
 		getter: customGetter,
 		fs:     &standardFS{},
 	}
+
+	// NewSession can fail. If metadata cannot be initialized the region string is populated with the default value
```
samuelkarp: Please remove "NewSession can fail."; it's redundant.

ecs-init/cache/cache_test.go
samuelkarp: I realize that the default path is the same as an unmapped region; I want that to be tested explicitly.  Please use a different return value than `config.DefaultRegionName`.

ecs-init/config/common.go
```diff
@@ -29,8 +29,21 @@ const (
...
+
+	// DefaultRegionName is the name of the region to fall back to if no entry for the region name is found in the
+	// S3BucketMap.
+	DefaultRegionName = "default"
 )
 
+// s3BucketMap provides a mapping of region names to specific URI's for the region.
+var s3BucketMap = map[string]string{
+	"cn-north-1" : "https://s3.cn-north-1.amazonaws.com.cn/amazon-ecs-agent/",
+	"default" : "https://s3.amazonaws.com/amazon-ecs-agent/",
```
samuelkarp: Use `DefaultRegionName` here instead of the literal `"default"`.

ecs-init/config/common.go
```diff
@@ -117,11 +122,14 @@ func DockerUnixSocket() (string, bool) {
 
 // getBaseLocationForRegion fetches the bucket URI from list of S3 Buckets by region name or default if key is not found
 func getBaseLocationForRegion(regionName string) string {
+	if regionName == RegionNameNotFound {
```
samuelkarp: Why do you need this here?  You'll have that below in the `if !ok` block.

jahkeup: There are 3 possible cases that are handled here:

1. regionName == RegionNameNotFound: Couldn't resolve your region for some reason
2. regionName == "us-east-2": Not explicitly in the map
3. regionName == "cn-north-1" (or some other map member): Key is in the map

Your [earlier comments](https://github.com/aws/amazon-ecs-init/pull/87#discussion_r119472709) led @cabbruzzese and me to see that the region resolving code was not separating the case of an unresolved region and a default region - which would have been intentionally passed to this function if desired - and thus we scoped the defaulting/fallback behavior to this function.

samuelkarp: That comment is specifically about the tests not covering the case, not whether the underlying code works.  I don't think it matters whether the code separates those two cases, but it does matter that both are covered by tests.  As long as we have test coverage that exercises the different situations we expect to have happen I'll be happy.

jahkeup: SGTM, I believe we have this covered now.

jahkeup: @samuelkarp can you confirm [this mocked response in TestDownloadAgentSuccessWithRegionFailure](https://github.com/aws/amazon-ecs-init/pull/87/files#diff-0be318ad582dbc3fdf05dca3c0e463bfR418) adequately models the failure you've mentioned?

samuelkarp: I would expect that to be `config.DefaultRegionName` based on the version of the code that didn't have this superfluous `RegionNameNotFound`.  But yes, that looks like it models the failure to resolve a region from the metadata service.

PR #94, opened 2017-05-24
Vendor upgrade

vsiddharth:
* Re-vendoring dependencies according to the new style. 
* Fixed test failures
* Updated scripts + Makefile


ecs-init/Godeps/Godeps.json
```diff
@@ -1,22 +1,240 @@
...
+			"Rev": "f561c5e57575bb1e0a2167028b7339b3a8d16fb4"
+		},
+		{
+			"ImportPath": "github.com/cihub/seelog/archive/zip",
+			"Comment": "v2.6-31-gf561c5e",
+			"Rev": "f561c5e57575bb1e0a2167028b7339b3a8d16fb4"
+		},
+		{
+			"ImportPath": "github.com/docker/docker/api/types",
+			"Comment": "docs-v1.12.0-rc4-2016-07-15-6275-g80013fd59",
```
samuelkarp: This looks like a really weird branch.  Can you check what you have in your GOPATH and see if we can get something like 17.03.1-ce here?

ecs-init/VERSION
```diff
@@ -1 +0,0 @@
-1.14.1
```
samuelkarp: Why is this removed?

vsiddharth: It was accidental. I'll restore it.

ecs-init/Godeps/Godeps.json
vsiddharth: This is something `godep` performed. 
Do we really need to manually check and update all the vendor dependencies ?

cabbruzzese: This file is being included by godeps as a dependency of go-dockerclient. I don't think we should try to change it.

@vsiddharth did you have an existing copy of github.com/docker/docker/ in your $GOPATH/src before using godep save? Wouldn't hurt to go get go-dockerclient again to make sure.

vsiddharth: @cabbruzzese I did not have an existing copy of `go-dockerclient` while updating these dependencies. I had a a fresh `go-workspace` when I tried this.

petderek: @samuelkarp  agent uses a similarly-named branch of docker, probably for the same reason.

vsiddharth:
@cabbruzzese @petderek Could you have another look at this PR ?
I've resolved the merge conflicts and made minor changes to a few build scripts. 

Thanks

ecs-init/Godeps/Godeps.json
samuelkarp: @vsiddharth @cabbruzzese @petderek We should figure out exactly why this is happening and fix it both here and in the agent.

PR #95, opened 2017-05-26
CAP_NET_ADMIN for amazon-ecs-agent

vsiddharth:
Adding `CAP_NET_ADMIN` for amazon-ecs-agent

Testing

- [x] make static
- [x] make gotest

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -53,6 +53,10 @@ const (
 	// maxRetries specifies the maximum number of retries for ping to return
 	// a successful response from the docker socket
 	maxRetries = 5
+	// CAP_NET_ADMIN to start agent
+	// For more information on capabilities, please read this manpage:
+	// http://man7.org/linux/man-pages/man7/capabilities.7.html
+	CAP_NET_ADMIN = "NET_ADMIN"
 )
 
 // Client enables business logic for running the Agent inside Docker
@@ -228,6 +232,7 @@ func (c *Client) getHostConfig() *godocker.HostConfig {
 		Binds:       binds,
 		NetworkMode: networkMode,
 		UsernsMode:  usernsMode,
+		CapAdd:      []string{CAP_NET_ADMIN},
 	}
 }
 
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -53,10 +53,10 @@ const (
 	// maxRetries specifies the maximum number of retries for ping to return
 	// a successful response from the docker socket
 	maxRetries = 5
-	// CAP_NET_ADMIN to start agent
+	// CapNetAdmin to start agent with NET_ADMIN capability
 	// For more information on capabilities, please read this manpage:
 	// http://man7.org/linux/man-pages/man7/capabilities.7.html
-	CAP_NET_ADMIN = "NET_ADMIN"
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker.go
```diff
@@ -53,6 +53,10 @@ const (
 	// maxRetries specifies the maximum number of retries for ping to return
 	// a successful response from the docker socket
 	maxRetries = 5
+	// CAP_NET_ADMIN to start agent
+	// For more information on capabilities, please read this manpage:
+	// http://man7.org/linux/man-pages/man7/capabilities.7.html
+	CAP_NET_ADMIN = "NET_ADMIN"
```
aaithal: Can you rename this to be a go style const? `CapNetAdmin`

ecs-init/docker/docker_test.go
```diff
@@ -253,6 +253,14 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
 	if hostCfg.NetworkMode != networkMode {
 		t.Errorf("Expected network mode to be %s, got %s", networkMode, hostCfg.NetworkMode)
 	}
+
+	if len(hostCfg.CapAdd) != 1 {
```
aaithal: Can we use `github.com/stretchr/testify/assert.Len()` instead?

ecs-init/docker/docker.go
vsiddharth: Sure

vsiddharth:
@aaithal On the subject of tests, we already have a number of tests that `Start`|`Stop` the agent and verify the attributes of the agent container. 

Are you suggesting a functional test here ?

ecs-init/docker/docker_test.go
vsiddharth: We can but none of the other tests in the file employ the assert package. 
We haven't even vendored this package for `ecs-init`. 

It though it would be strange to have two styles in the same file.

aaithal: Ah, alright then. I think we should vendor that for this package. May be not for this PR, but it'd be nice to start using some `assert`'s here.

vsiddharth: We can refactor the tests and make them use the assert package. 
We can track this as a new [issue](https://github.com/aws/amazon-ecs-init/issues/98)

aaithal:
@vsiddharth if we can add one, then yes. I was mostly asking for a manual launch and verification.

vsiddharth:
@aaithal I have manually tested and verified that this change starts the agent with `NET_ADMIN` capability.

vsiddharth:
docker inspect $AGENT_CID
```
[
    {
        "Id": "10e0792661f08113df02fe15ed815663043bd7faf5566b084ac52fb81ccad267",
        "Created": "2017-05-30T21:27:32.20915394Z",
        "Path": "/agent",
        "Args": [],
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 6032,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2017-05-30T21:27:32.667930338Z",
            "FinishedAt": "0001-01-01T00:00:00Z"
        },
...
"CapAdd": [
         "NET_ADMIN"
 ],
"CapDrop": null,
...
```

PR #96, opened 2017-05-26
update to support agent 1.14.2

adnxn:
* Cache Agent version 1.14.2


nmeyerhans:
@aaithal Nope, that file no longer needs to change. Values are injected at runtime as of PR #91

aaithal:
@nmeyerhans oh cool, nice.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -149,6 +149,9 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Fri May 26 2017 Adnan Khan <adnkha@amazon.com> - 1.14.2-1
+- Cache Agent version 1.14.2
+- Add functionality for region based agent updates
```
nmeyerhans: There should be an entry here about running agent with userns=host when Docker is running with userns-remap enabled.

adnxn: gotcha. also i just realised it shouldn't have that region based agent updates entry for this release.

nmeyerhans:
:ship: LGTM

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -149,6 +149,9 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Thu Jun 1 2017 Adnan Khan <adnkha@amazon.com> - 1.14.2-1
+- Cache Agent version 1.14.2
+- Add functionality for running agent with userns=host when Docker has userns-remap enabled
```
aaithal: I think we're missing an entry here and in the spec file for the docker `17.03.x-ce` upgrade.

samuelkarp: @aaithal is correct, see line 30.

adnxn: fixed both things.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -149,6 +149,10 @@ if [ -e %{running_semaphore} ]; then
...
+- Add support for Docker 17.03.1ce
```
aaithal: Should we say `ce` here if that's not specified in the spec file?

adnxn: i'm not sure tbh. i figured if the user checks ```docker version``` they would get the "-ce" bit as well.

aaithal: let's be consistent. If we're not calling it 'ce' in the spec file, we should leave that bit here too i think.

adnxn: turns out we *have* to call it "ce" in the spec file, otherwise it makes rpm versioning sad.

PR #99, opened 2017-06-02
Mount /proc for agent

vsiddharth:
Mount `/proc` 

**Description**
With task-enis, the ECS agent needs access to the container's
network namespace and for this; the agent needs access to the host procfs. This
patch mounts the host `/proc` as `/host/proc` within the agent container.

**Testing**
```
"HostConfig": {
            "Binds": [
                "/var/run:/var/run",
                "/var/log/ecs:/log",
                "/var/lib/ecs/data:/data",
                "/etc/ecs:/etc/ecs",
                "/var/cache/ecs:/var/cache/ecs",
                "/proc:/host/proc"
            ],


"Mounts": [
        ….     
	     {
                "Source": "/proc",
                "Destination": "/host/proc",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
	     },
	….
 ],
```

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -29,6 +29,9 @@ const (
 	AgentLogFile = "ecs-agent.log"
 
 	UnixSocketPrefix = "unix://"
+
+	// Used to mount /proc for agent container
+	ProcFS = "/proc"
 )
 
 // AgentConfigDirectory returns the location on disk for configuration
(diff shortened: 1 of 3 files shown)


jhaynes:
Can you add a bit of description and justification to the PR description? I don't have much context for why we're making this change.

Separately, the whole `/proc` dir is a big place. Do we really need the whole thing? If so, why?

PR #100, opened 2017-06-09
ubuntu trusty: Updated to Go 1.7

rickard-von-essen:
Also bumped version to 1.4.2-2

Requires PPA version of Go since 1.6 is the latest for Trusty. ~~For example [this](https://launchpad.net/~jonathonf/+archive/ubuntu/golang?field.series_filter=trusty).~~ [Go 1.X packages](https://launchpad.net/%7Egophers/+archive/ubuntu/archive?field.series_filter=trusty)  from The Go Language Gophers” team.

### DIFF
--- a/packaging/ubuntu-trusty/debian/README.source
+++ b/packaging/ubuntu-trusty/debian/README.source
@@ -1,0 +1,7 @@
+This package requires Go 1.7 which is only availible for Ubuntu Trusty from PPA.
+
+  * Install Go 1.7 (golang-1.7) from ppa:gophers/archive, see
+    https://launchpad.net/%7Egophers/+archive/ubuntu/archive?field.series_filter=trusty
+
+      add-apt-repository ppa:gophers/archive
+      apt-get install golang-1.7
--- a/packaging/ubuntu-trusty/debian/changelog
+++ b/packaging/ubuntu-trusty/debian/changelog
@@ -1,5 +1,6 @@
 amazon-ecs-init (1.14.3-1) trusty; urgency=low
 
   * Initial release
+    Requires Go 1.7 (golang-1.7) from ppa see README.source
 
  -- Samuel Karp <skarp@amazon.com>  Wed, 22 Feb 2017 10:03:16 +0000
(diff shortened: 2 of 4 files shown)


samuelkarp:
Thanks for sending this!  We'll want to hold off on merging this until after https://github.com/aws/amazon-ecs-init/pull/101 and v1.14.3 of the agent is released (so it'd also make sense to update this to 1.14.3 and skip 1.14.2.).  For more about why we're getting 1.14.3 out quickly, see https://github.com/aws/amazon-ecs-agent/issues/833.

Any reason you'd lean toward [this PPA](https://launchpad.net/%7Ejonathonf/+archive/ubuntu/golang?field.series_filter=trusty) over [this one](https://launchpad.net/~gophers/+archive/ubuntu/archive?field.series_filter=trusty)?  If we're requiring a PPA to build, it'd probably be good to add that to the docs somewhere and to the `debian/changelog` file.

rickard-von-essen:
> Any reason you'd lean toward this PPA over this one?

No just sloppy searching.

> requiring a PPA to build, it'd probably be good to add that to the docs

I'll add a note.

I'll updated this as soon as v1.14.3 is out.

nmeyerhans:
Although we're not by any means bound by (or generally in compliance with) the Debian policy, it'd be nice for changelog to actually be an actual log rather than the stub we have here.

Additionally I'd prefer to see the Golang 1.7 PPA requirement documented in README.source and not just in the changelog. It'll improve discoverability for people who may just be getting started trying to build this package.

rickard-von-essen:
> it'd be nice for changelog to actually be an actual log rather than the stub we have here.

Agree. My hope when I refurbished @samuelkarp's initial PR to add Ubuntu Trusty packaging was that AWS would publish (unofficial) packages for this and release it at the same time as those for Amazon Linux.

rickard-von-essen:
It would be nice if you would provide builds for Ubuntu Trusty. Since that is to my knowledge the only other major server distro using upstart still, until its end of life April 2019.

rickard-von-essen:
Updated for 1.14.3-1 and added `README.source`.

nmeyerhans:
Merged this into the dev branch. It'll be included in future releases.

PR #103, opened 2017-06-16
Reenabled Lint script

cabbruzzese:
Updated lint script ignore section to use new vendor folder. Enabled script due to increase in go version.


aaithal:
when do we run this?

jhaynes:
It runs via `make test` or `make lint`.

PR #109, opened 2017-08-10
Add a systemd unit file for running ECS Agent

nmeyerhans:
This change introduces a systemd unit file that configures and runs the agent in much the same way as the golang ecs-init code. It is intended to be used in place ecs-init on systemd-based systems, although it is not currently a complete replacement for init.

 * The unit file does an unconditional pull of the latest ECS Agent from Docker Hub at startup. Neither the "unconditional" part nor the "Docker Hub" part are desirable.
 * The unit file does not currently handle the Agent's self-upgrade functionality. It may require changes to the agent's behavior in order to properly support this here.

The unit file is stored in a new "integration" subdirectory as there didn't seem to be a better place to store it. In theory this directory can contain additional configuration or scripts to handle integration with other systems if people want to add support for e.g. `runit` or `openrc`.

I have tested this on both Debian 9 (stretch) and Fedora 25. It works without modification on both.

### DIFF
--- a/integration/systemd/ecs-agent.service
+++ b/integration/systemd/ecs-agent.service
@@ -30,7 +30,7 @@ After=docker.service
 WantedBy=multi-user.target
 
 [Service]
-Environment=AGENT_TAG=v1.16.0
+Environment=AGENT_TAG=v1.16.1
 Type=simple
 # Load an updated agent, if it exists:
 ExecStartPre=-/bin/sh -c 'test -f /var/cache/ecs/desired-image && docker load $(cat /var/cache/ecs/desired-image) && rm -f $(cat /var/cache/ecs/desired-image) /var/cache/ecs/desired-image'
--- a/integration/systemd/ecs-agent.service
+++ b/integration/systemd/ecs-agent.service
@@ -30,7 +30,7 @@ After=docker.service
 WantedBy=multi-user.target
 
 [Service]
-Environment=AGENT_TAG=v1.16.1
+Environment=AGENT_TAG=v1.16.2
 Type=simple
 # Load an updated agent, if it exists:
 ExecStartPre=-/bin/sh -c 'test -f /var/cache/ecs/desired-image && docker load $(cat /var/cache/ecs/desired-image) && rm -f $(cat /var/cache/ecs/desired-image) /var/cache/ecs/desired-image'
--- a/integration/systemd/ecs-agent.service
+++ b/integration/systemd/ecs-agent.service
@@ -27,8 +27,12 @@ Requires=docker.service
 After=docker.service
 
 [Service]
+Environment=AGENT_TAG=v1.15.0
 Type=simple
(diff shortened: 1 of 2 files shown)


jgrumboe:
+1

rickard-von-essen:
See https://github.com/aws/amazon-ecs-agent/pull/1140

integration/systemd/ecs-agent.service
```diff
@@ -0,0 +1,81 @@
...
+  --net=host \
+  --env ECS_LOGFILE=/log/ecs-agent.log \
+  --env ECS_DATADIR=/data \
+  --env ECS_UPDATES_ENABLED=false \
+  --env ECS_AVAILABLE_LOGGING_DRIVERS='["json-file","syslog","awslogs"]' \
+  --env ECS_ENABLE_TASK_IAM_ROLE=true \
+  --env ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST=true \
+  --env ECS_UPDATES_ENABLED=true \
+  --env ECS_ENABLE_TASK_ENI=true \
+  --env-file=/etc/ecs/ecs.config \
```
aaithal: I think we need an `ExecStartPre` hook to create the `/etc/ecs` directory and to touch the `/etc/ecs/ecs.config` file as well.

integration/systemd/load-and-verify-agent.sh
```diff
@@ -0,0 +1,43 @@
...
+#     http://aws.amazon.com/apache2.0/
+# 
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# This wrapper script ensures that we've loaded the desired ECS agent
+# and have attached the "latest" tag to it.
+# USAGE: load-and-verify-agent.sh DEFAULT_TAG
```
aaithal: Just for my own understanding, when/where is this file supposed to be used?

nmeyerhans: The intent was to make this part of the agent in-place upgrade workflow. It's not used in the unit file currently because I wasn't able to come up with the right way to represent the existing workflow in systemd.

integration/systemd/ecs-agent.service
nmeyerhans: Yes, that's reasonable. I hadn't been doing it in the unit file in my testing, but was instead populating the ecs.config file via userdata. However, in the case where userdata doesn't generate such a file, we should still ensure that it exists.

reecejmarkowsky:
+1

sparrc:
this is largely out of date with what ecs-init is currently doing. It would be nice to have this functionality, but I dont think it's something we want to officially support in parallel to amazon-ecs-init.

PR #110, opened 2017-08-22
update to support agent 1.14.4

jhaynes:
- Cache Agent version 1.14.4
- update Ubuntu packaging for Trusty


packaging/ubuntu-trusty/debian/changelog
```diff
@@ -1,5 +1,6 @@
-amazon-ecs-init (1.14.0-2) trusty; urgency=low
+amazon-ecs-init (1.14.3-1) trusty; urgency=low
```
nmeyerhans: If we're going to ship this file, we should probably update it as we release. Otherwise people who build packages from it will get confusing results.

ecs-init/version/version.go
```diff
@@ -18,10 +18,10 @@
 package version
 
 // Version is the version of the ecs-init
-var Version string = "1.14.3"
+var Version string = "1.14.4"
 
 // GitDirty indicates the cleanliness of the git repo when this ecs-init was built
-var GitDirty string = "true"
+var GitDirty string = "false"
```
samuelkarp: This should not be changed.

PR #112, opened 2017-08-30
Start agent with task eni enabled flag

aaithal:
### Summary
Changes required to start the Agent with Task ENI flag/feature enabled

### Details
* 555ca37 ecs-init/docker,packaging/amazon-linux-ami: start agent with --init
* 393b051 packaging/amazon-linux-ami: modify spec file with dhclient dependencies
* 7ccbab5 ecs-init/docker,ecs-init/config: modify agent create options to support task enis

### Testing Done
* [X] build (`make clean rpm`) succeeds
* [X] tests (`make gotest`) succeed
* [X] manual tests (installed the RPM and started the ECS Agent and verified that it registers with the `task-eni` capability)

### DIFF
--- a/ecs-init/Godeps/Godeps.json
+++ b/ecs-init/Godeps/Godeps.json
@@ -305,7 +305,8 @@
 		},
 		{
 			"ImportPath": "github.com/fsouza/go-dockerclient",
-			"Rev": "c933ed18bef34ec2955de03de8ef9a3bb996e3df"
+			"Comment": "docker-1.9/go-1.4-421-g98edf3e",
+			"Rev": "98edf3edfae6a6500fecc69d2bcccf1302544004"
 		},
 		{
 			"ImportPath": "github.com/go-ini/ini",
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -26,13 +26,32 @@ import (
 )
 
 const (
-	logDir      = "/log"
-	dataDir     = "/data"
-	readOnly    = ":ro"
+	// logDir specifies the location of Agent log files in the container
+	logDir = "/log"
+	// dataDir specifies the location of Agent state file in the container
+	dataDir = "/data"
+	// readOnly specifies the read-only suffix for mounting host volumes
+	// when creating the Agent container
+	readOnly = ":ro"
+	// hostProcDir binds the host's /proc directory to /host/proc within the
+	// ECS Agent container
+	// The ECS Agent needs access to host's /proc directory when configuring
+	// the network namespace of containers for tasks that are configured
+	// with an ENI
 	hostProcDir = "/host/proc"
-	// set default to /var/run instead of /var/run/docker.sock in case
-	// /var/run/docker.sock is deleted and recreated outside the container
+	// defaultDockerEndpoint is set to /var/run instead of /var/run/docker.sock
+	// in case /var/run/docker.sock is deleted and recreated outside the container
 	defaultDockerEndpoint = "/var/run"
+	// dhclientLeasesLocation specifies the location where dhclient leases
+	// information is tracked in the Agent container
+	dhclientLeasesLocation = "/var/lib/dhclient"
+	// dhclientLibDir specifies the location of shared libraries on the
+	// host and in the Agent container required for the execution of the dhclient
+	// executable
+	dhclientLibDir = "/lib64"
+	// dhclientExecutableDir specifies the location of the dhclient
+	// executable on the  host and in the Agent container
+	dhclientExecutableDir = "/sbin"
 	// networkMode specifies the networkmode to create the agent container
 	networkMode = "host"
 	// usernsMode specifies the userns mode to create the agent container
@@ -181,6 +200,7 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		"ECS_AVAILABLE_LOGGING_DRIVERS":         `["json-file","syslog","awslogs"]`,
 		"ECS_ENABLE_TASK_IAM_ROLE":              "true",
 		"ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST": "true",
+		"ECS_ENABLE_TASK_ENI":                   "true",
 	}
 
 	// merge in user-supplied environment variables
@@ -232,7 +252,10 @@ func (c *Client) getHostConfig() *godocker.HostConfig {
 		config.AgentDataDirectory() + ":" + dataDir,
 		config.AgentConfigDirectory() + ":" + config.AgentConfigDirectory(),
 		config.CacheDirectory() + ":" + config.CacheDirectory(),
-		config.ProcFS + ":" + hostProcDir,
+		config.ProcFS + ":" + hostProcDir + readOnly,
+		config.AgentDHClientLeasesDirectory() + ":" + dhclientLeasesLocation,
+		dhclientLibDir + ":" + dhclientLibDir + readOnly,
+		dhclientExecutableDir + ":" + dhclientExecutableDir + readOnly,
 	}
 	return &godocker.HostConfig{
 		Binds:       binds,
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -262,6 +262,7 @@ func (c *Client) getHostConfig() *godocker.HostConfig {
 		NetworkMode: networkMode,
 		UsernsMode:  usernsMode,
 		CapAdd:      []string{CapNetAdmin, CapSysAdmin},
+		Init:        true,
 	}
 }
 
--- a/packaging/amazon-linux-ami/ecs-init.spec
+++ b/packaging/amazon-linux-ami/ecs-init.spec
@@ -31,7 +31,7 @@ Requires:       docker >= 1.6.0, docker <= 17.03.2ce
 Requires:       upstart
 Requires:       iptables
 Requires:       procps
-Requires(post): docker >= 1.6.0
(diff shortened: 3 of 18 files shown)


ecs-init/Godeps/Godeps.json
```diff
@@ -305,7 +305,8 @@
 		},
 		{
 			"ImportPath": "github.com/fsouza/go-dockerclient",
-			"Rev": "c933ed18bef34ec2955de03de8ef9a3bb996e3df"
+			"Comment": "docker-1.9/go-1.4-421-g98edf3e",
```
samuelkarp: This looks like a very weird tag.

ecs-init/docker/docker.go
```diff
@@ -26,13 +26,32 @@ import (
...
 	hostProcDir = "/host/proc"
-	// set default to /var/run instead of /var/run/docker.sock in case
-	// /var/run/docker.sock is deleted and recreated outside the container
+	// defaultDockerEndpoint is set to /var/run instead of /var/run/docker.sock
+	// in case /var/run/docker.sock is deleted and recreated outside the container
 	defaultDockerEndpoint = "/var/run"
+	// dhclientLeasesLocation specifies the location where dhclient leases
+	// information is tracked in the Agent container
+	dhclientLeasesLocation = "/var/lib/dhclient"
+	// dhclientLibDir specifies the location of shared libraries on the
```
samuelkarp: Is there a dependency on a specific version of libc here?  Are you mounting in the dhclient executable from the host?

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -27,11 +27,11 @@ Source1:        ecs.conf
 
 BuildRequires:  golang >= 1.7
 
-Requires:       docker >= 1.6.0, docker <= 17.03.2ce
+Requires:       docker = 17.03.2ce
 Requires:       upstart
 Requires:       iptables
 Requires:       procps
-Requires(post): docker >= 1.6.0
+Requires:       dhclient
```
samuelkarp: We'll want to update both our SUSE and Ubuntu packages as well to have the new dependency.

@rjschwei I assume that SUSE has dhclient in a package; can you confirm that this will be a requirement that can be satisfied on SUSE?

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -27,11 +27,11 @@ Source1:        ecs.conf
 
 BuildRequires:  golang >= 1.7
 
-Requires:       docker >= 1.6.0, docker <= 17.03.2ce
+Requires:       docker = 17.03.2ce
```
samuelkarp: We'll want to update both our SUSE and Ubuntu packages as well to have the new version.

@rjschwei Does SUSE have a new-enough version of Docker?  I think that 1.13 is the earliest we'd be able to work with, as long as `--init` functions properly.

ecs-init/docker/docker.go
aaithal: We're mounting `libc` from the host.

packaging/amazon-linux-ami/ecs-init.spec
aaithal: Sure, but i'm keeping that as out of scope for this pr

ecs-init/docker/docker.go
rjschwei: That location does not exists in SLES. SLES uses wicked for networking and leases stored in a different location in XML format, for example:

# cat /var/lib/wicked/lease-eth0-dhcp-ipv4.xml
<lease>
  <family>ipv4</family>
  <type>dhcp</type>
  <uuid>b3bda659-d649-0800-e403-000004000000</uuid>
  <state>granted</state>
  <update>0x00000000</update>
  <acquired>1504272564</acquired>
  <ipv4:dhcp>
    <client-id>01:42:01:0a:f0:00:06</client-id>
    <server-id>169.254.169.254</server-id>
    <relay-address>10.240.0.1</relay-address>
    <sender-hw-address>42:01:0a:f0:00:01</sender-hw-address>
    <lease-time>86400</lease-time>
    <renewal-time>43200</renewal-time>
    <rebind-time>75600</rebind-time>
    <hostname>sles12devtestguest.c.suse-gce-test.internal</hostname>
    <address>10.240.0.6</address>
    <netmask>255.255.255.255</netmask>
    <mtu>1460</mtu>
    <boot>
      <server-address>10.240.0.1</server-address>
    </boot>
    <routes>
      <route>
        <destination>10.240.0.1/32</destination>
      </route>
      <route>
        <nexthop>
          <gateway>10.240.0.1</gateway>
        </nexthop>
      </route>
    </routes>
    <dns>
      <domain>c.suse-gce-test.internal</domain>
      <server>169.254.169.254</server>
      <search>c.suse-gce-test.internal</search>
      <search>google.internal</search>
    </dns>
    <ntp>
      <server>169.254.169.254</server>
    </ntp>
  </ipv4:dhcp>
</lease>

packaging/amazon-linux-ami/ecs-init.spec
rjschwei: Yes, current version of docker on SLES is 1.8.2 and 1.10.2 is in the pipeline.

packaging/amazon-linux-ami/ecs-init.spec
rjschwei: No, for SLES the network handling is done with wicked, so the dhcp requirement can be met with a different implementation, however the apparent dependence on the format and location of the lease information file is going to be a problem.

ecs-init/docker/docker.go
aaithal: @rjschwei can you please let me know which location are you referring to? If you're referring to `dhclientLeasesLocation = "/var/lib/dhclient"`, that's something that need not be present on the host. It's a volume mount target within the Agent container.

rjschwei: @aaithal well dhclientLeasesLocation = "/var/lib/dhclient" needs to be present now, but it is not present on SLES and will not be, basically this update breaks SUSE support.

Leas information is located in /var/lib/wicked in SLES and the file is in a different format than the dhclient lease file

aaithal: @rjschwei It seems there's a bit of disconnect here and I apologize for that. Let me try to provide some details here. The `dhclientLeasesLocation = "/var/lib/dhclient"` specifies the location where the ECS Agent will write lease information to. Since the ECS Agent will be launched as a container, it can have it's own filesystem structure that's different than that of the host. So, the ECS Agent container will write leases file in this location. The way we're doing this for Amazon Linux is by creating the `"/var/lib/ecs/dhclient"` on the host (you can find a refernece to this here: https://github.com/aws/amazon-ecs-init/blob/enis/ecs-init/config/common.go#L81) via the `ecs-init.spec` file in https://github.com/aws/amazon-ecs-init/blob/enis/packaging/amazon-linux-ami/ecs-init.spec#L79.

Essentially, ECS Agent will be launched with `-v /var/lib/ecs/dhclient:/var/lib/dhclient`. The host part of this mount will be created via the RPM spec file. I just had a brief look at the spec file for suse and it seems that the install section can be modified to create this on suse as well: https://github.com/aws/amazon-ecs-init/blob/enis/packaging/suse/amazon-ecs-init.spec#L48 (please correct me if i'm wrong).

packaging/amazon-linux-ami/ecs-init.spec
nmeyerhans: Is dhclient available at all in SLES? If so, it'd be easiest to let it come in as a dependency here. The corresponding [amazon-ecs-agent work](https://github.com/aws/amazon-ecs-agent/tree/enis) is fairly tightly coupled to dhclient right now.

Using dhclient for the ENI support wouldn't require the use of dhclient for system-wide network configuration, assuming the two packages can be installed side-by-side. There may be additional work required to ensure they don't step on each other when task ENIs are attached.

rjschwei: No, dhclient is not available for SLES, not packaged and supported by SUSE anyway.

ecs-init/docker/docker.go
rjschwei: Well, yes we can create any path via a spec file thus on the host we can certainly create /var/lib/dhclient and if that information is not actually needed to be shared with the host than there is no problem. However, if the host for some reason or another needs to be aware of the lease information then that is not going to work on SLES.

Further, if the agent moves into a container from running directly on the host then somehow that container has to be pulled onto the host, and I obviously do not follow the work here  closely enough to know how that is being resolved. Anyway, if the agent runs in a container and you provide the container then of course any dhcp handler of choice can be used. However, if the expectation is that I build a container that contains the agent and then build that container into the VM image and launch it on start up then the use of dhclient is a problem.

packaging/amazon-linux-ami/ecs-init.spec
samuelkarp: @rjschwei Sounds like SLES does *not* have a new-enough version of Docker.  We may want to start with `--pid=host` on SLES instead of `--init`.

rjschwei: Ah, sorry I didn't read that correctly, but with SLES 12 SP3, to be released on Thursday:

~> rpm -qa docker
docker-17.04.0_ce-98.2.x86_64

So going forward we are OK, given that the SP2 images will be replaced by SP3, although SP3 will still have version 1.13.0 I don't think we need anything special :smile:

PR #115, opened 2017-09-05
Disable task enis on suse/ubuntu

aaithal:
### Summary
disable task eni feature on suse and ubuntu

### Details
Respect build flags when building Agent container config and host config to disable the Task ENI feature on suse and ubuntu platforms

### Testing Done
* [X] `make clean rpm` 
* [X] `make test-in-docker`
* [X]  `./scripts/gobuild.sh suse`
* [X]  `./scripts/gobuild.sh ubuntu`
* [X]  Manual testing: Verified that `ECS_ENABLE_TASK_ENI=true` is set only when Init is built for Amazon Linux and disabled when built with `suse` and `ubuntu` tags by inspecting the Agent container

### DIFF
--- a/ecs-init/docker/dependencies.go
+++ b/ecs-init/docker/dependencies.go
@@ -28,15 +28,6 @@ import (
 	godocker "github.com/fsouza/go-dockerclient"
 )
 
-const (
-	// dockerClientAPIVersion specifies the minimum docker client API version
-	// required by ECS Init
-	// Version 1.25 is required for setting Init to true when constructing
-	// the HostConfig for creating the ECS Agent to enable the Task
-	// Networking with ENI capability
-	dockerClientAPIVersion = "1.25"
-)
-
 type dockerclient interface {
 	ListImages(opts godocker.ListImagesOptions) ([]godocker.APIImages, error)
 	LoadImage(opts godocker.LoadImageOptions) error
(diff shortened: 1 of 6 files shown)


aaithal:
@samuelkarp good point. I've modified the api client version as well.

samuelkarp:
@rjschwei Can you take a look here too?

rjschwei:
Sorry a bit slow, in the middle of a release, I don't see anything obvious, what is eni?

aaithal:
@rjschwei ENI refers to [Amazon EC2 Elastic Network Interface](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html). We are working on a feature where by containers in a task are provisioned with an ENI when launched with a new network mode option. You can find the proposal [here](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html)

rjschwei:
OK, I thought it maybe, but better to as as this is referred to as ENA (Elastic Network Adapter) support. Anyway, SLES support ENI/ENA and thus the feature should NOT be disabled on SLES.

Thanks

aaithal:
@rjschwei we have a dependency on `dhclient` executable (the CNI plugin does) in the current implementation, which we are mounting from the host into the Agent container, because of which we're disabling that feature (Provisioning ENIs for ECS Tasks) on SLES at launch. However, we'll revisit this decision after launch.

Thanks,
Anirudh

PR #120, opened 2017-10-16
Task cgroup mountpoint

petderek:
Adding config to detect where cgroups are located on host. This change will inspect two mountpoints (/sys/fs/cgroup and /cgroup) to determine the correct placement. It will then pass that value into agent.


ecs-init/config/common.go
```diff
@@ -133,3 +137,15 @@ func getBaseLocationForRegion(regionName string) string {
 
 	return s3BucketURL
 }
+
+// GetCgroupMountpoint returns the cgroup mountpoint
+func GetCgroupMountpoint() (string, error) {
```
jahkeup: Is this resolution of the mountpoint outside or inside the container here? Should we call this something like `GetHostCgroupPath`?

ecs-init/docker/docker.go
```diff
@@ -173,10 +173,14 @@ func (c *Client) findAgentContainer() (string, error) {
 
 // StartAgent starts the Agent in Docker and returns the exit code from the container
 func (c *Client) StartAgent() (int, error) {
+	hostConfig, err := c.getHostConfig()
+	if err != nil {
+		return 0, err
```
jahkeup: Should we have a different error code when the agent can't be configured? This failure mode seems different to a failure starting the container IMO.

ecs-init/docker/docker_test.go
```diff
@@ -234,7 +234,7 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
 
 	hostCfg := opts.HostConfig
 
-	if len(hostCfg.Binds) != 5 {
+	if len(hostCfg.Binds) != 6 {
 		t.Errorf("Expected exactly 5 elements to be in Binds, but was %d", len(hostCfg.Binds))
```
jahkeup: s/5/6/

ecs-init/config/common.go
petderek: The resolution is outside the container, so I think the current name makes sense.

ecs-init/docker/docker_test.go
petderek: Good catch, will fix

ecs-init/docker/docker.go
petderek: The other thing I'm wondering is should we log an error and continue? Is this worth blocking startup?

samuelkarp:
:-1:

I don't think we should take the approach of attempting to detect the location.  Instead, I think we should hard-code the location for the known Linux distributions where we have builds: `/cgroup` for Amazon Linux and `/sys/fs/cgroup` for SUSE and Ubuntu.  We already specify build flags for this (see [this file](https://github.com/aws/amazon-ecs-init/blob/bef8d94863624c3d045d264d873a72c934b43e74/ecs-init/docker/dependencies_suse_ubuntu.go#L1) and [this file](https://github.com/aws/amazon-ecs-init/blob/bef8d94863624c3d045d264d873a72c934b43e74/ecs-init/docker/dependencies_unspecified.go#L1)) to encode distro-specific configuration.

petderek:
@samuelkarp @vsiddharth better? 😄

ecs-init/config/common.go
```diff
@@ -124,6 +124,15 @@ func DockerUnixSocket() (string, bool) {
 	return "/var/run", false
 }
 
+// CgroupMountpoint returns either the default cgroup mountpoint for the system or the mountpoint overidden at CGROUP_MOUNTPOINT
+func CgroupMountpoint() string {
+	if mount := os.Getenv("CGROUP_MOUNTPOINT"); mount != "" {
```
samuelkarp: Do we need an override?

We've talked a bit about configuring ecs-init before and erred on the side of not allowing configuration.  This makes ecs-init easier to understand (fewer possibilities).  If you're just using this for development, there's actually a [`development` build tag](https://github.com/aws/amazon-ecs-init/blob/bef8d94863624c3d045d264d873a72c934b43e74/ecs-init/config/development.go#L1) that can be used and helps separate development use-cases from production use-cases.

ecs-init/config/common_test.go
```diff
@@ -66,3 +66,13 @@ func TestGetS3BucketMapByRegion(t *testing.T) {
 		})
 	}
 }
+
+func TestCgroupMountOverride(t *testing.T) {
+	cgroupTest := "/ecs/cgroup"
+	os.Setenv("CGROUP_MOUNTPOINT", cgroupTest)
```
samuelkarp: Any time you're setting an environment variable in a test, you should `defer os.Unsetenv`.

ecs-init/config/common.go
petderek: We don't *need* it. Its just technically possible to override where cgroups live on the host system. I'd rather include the option.

samuelkarp: My preference would be to exclude it for now; we can always add it again later, but it's hard to remove later if we don't know whether people are using it or not.  When we do add configuration to ecs-init, I want to think a little harder about whether we should use environment variables or provide a different configuration mechanism.

ecs-init/config/common.go
```diff
@@ -124,6 +124,11 @@ func DockerUnixSocket() (string, bool) {
 	return "/var/run", false
 }
 
+// CgroupMountpoint returns either the default cgroup mountpoint for the system or the mountpoint overidden at CGROUP_MOUNTPOINT
```
jahkeup: I think this should be updated now that we aren't going to have env override support.

ecs-init/config/config_unspecified.go
```diff
@@ -0,0 +1,18 @@
...
+// http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+const cgroupMountpoint = "/cgroup"
```
vsiddharth: I don't know on what other distro's this is true. 
I'm only aware of this on `Amazon Linux`. 

`CoreOS` for example uses `/sys/fs/cgroup`

petderek: I'm following the convention elsewhere in the package, where we default to Amazon Linux and have overrides in place for Suse and Ubuntu. 

We don't package ecs-init for CoreOS .

ecs-init/config/common.go
petderek: Agreed!

ecs-init/config/config_unspecified.go
vsiddharth: I'd think that `/cgroup` is rather specific only for `Amazon Linux` as most traditional distro's use `/sys/fs/cgroup`. 

I'd rather have a specific override for `Amazon Linux` and let the other unspecified ones default to `/sys/fs/cgroup`. 

I was not implying that we package things for `CoreOS`, just used it as an example as they have come back and reported issues with the `amazon-ecs-agent`.

samuelkarp: Right now we default to assuming Amazon Linux and treat other distributions as special-cases.  I think it's reasonable to do that for this right now.

petderek: It is specific for Amazon Linux -- but that's our default. Our build tools assume its for Amazon Linux unless otherwise specified via platform specific target.

vsiddharth: Thanks a lot for all the explanations, the file name `config_unspecified.go` and the lack of comments made it confusing for me.

ecs-init/docker/docker.go
```diff
@@ -273,3 +278,18 @@ func (c *Client) StopAgent() error {
 	stopContainerTimeoutSeconds := uint(10)
 	return c.docker.StopContainer(id, stopContainerTimeoutSeconds)
 }
+
+// getPlatformSpecificEnvVariables gets a map of environment variable key-value
```
samuelkarp: Can you move this back?  I believe this was intended for the task networking changes, which will only be enabled on Amazon Linux right now.

petderek: I had moved it because it was literally the same on both platforms. Is this changing?

samuelkarp: @aaithal would know better than me.  I think that's why it's the way it is.

aaithal: It's intended to be changed in the next init release. Please move it back.

samuelkarp:
Can you fix the merge conflicts?

jahkeup:
@petderek @vsiddharth are we blocked on merging this any longer?

vsiddharth:
@jahkeup This should be good to go!

PR #121, opened 2017-10-17
Update to support agent 1.15.0.

jhaynes:
* Update to support agent 1.15.0.


ecs-init/version/version.go
```diff
@@ -18,10 +18,10 @@
 package version
 
 // Version is the version of the ecs-init
-var Version string = "1.14.5"
+var Version string = "1.15.0"
 
 // GitDirty indicates the cleanliness of the git repo when this ecs-init was built
-var GitDirty string = "true"
+var GitDirty string = "false"
```
samuelkarp: This should be "true" any time it's checked in.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 Name:           ecs-init
-Version:        1.14.5
-Release:        1%{?dist}
+Version:        1.15.0
+Release:        2%{?dist}
```
aaithal: why is this `2`?

PR #128, opened 2017-11-08
add support for awslogs execution role

adnxn:
This change introduces a new `ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE` environment variable to enable the corresponding agent feature. Also increment the required docker version to `17.06.2ce`. The docker version bump is required since it includes the awslogs credentials endpoint flag introduced here https://github.com/moby/moby/pull/35055

### DIFF
--- a/ecs-init/docker/docker_test.go
+++ b/ecs-init/docker/docker_test.go
@@ -229,6 +229,7 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions, 
 	expectKey("ECS_ENABLE_TASK_IAM_ROLE=true", envVariables, t)
 	expectKey("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST=true", envVariables, t)
 	expectKey("ECS_ENABLE_TASK_ENI=true", envVariables, t)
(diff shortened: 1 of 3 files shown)


ecs-init/docker/docker_test.go
```diff
@@ -229,6 +229,7 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
 	expectKey("ECS_ENABLE_TASK_IAM_ROLE=true", envVariables, t)
 	expectKey("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST=true", envVariables, t)
 	expectKey("ECS_ENABLE_TASK_ENI=true", envVariables, t)
+	expectKey("ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE=true", envVariables, t)
```
samuelkarp: We should have a test that verifies that this is *not* set when not running Amazon Linux.

adnxn: > a test that verifies that this is not set when not running Amazon Linux.

sure. what would be the best way to test that?

adnxn: tracking in #131

PR #130, opened 2017-11-15
Version 1.15.2-2

nmeyerhans:
Update for agent version 1.15.2


ecs-init/version/version.go
```diff
@@ -18,7 +18,7 @@
 package version
 
 // Version is the version of the ecs-init
-var Version string = "unknown"
+var Version string = "1.15.2"
```
jhaynes: Since this is set at build time similarly to the `GitShortHash`, I think we should leave it as `unknown` also.

jhaynes:
It looks like the `AgentFilename` edit from `ecs-init/config/common.go` is missing from the diff currently.

nmeyerhans:
> It looks like the AgentFilename edit from ecs-init/config/common.go is missing from the diff currently.

Well isn't that unfortunate. Thanks for catching this.

PR #133, opened 2017-11-21
Correct typo Serivce > Service

endofcake:
Hello folks,

Just correcting an annoying typo here, nothing too onerous.


nmeyerhans:
Thanks for catching that. Merged.

Issue #92, opened 2017-05-19
ecs-init fails when it can't connect to docker


thesmith:
I believe this is a timing issue. I am using the ElasticBeanstalk-blessed image aws-elasticbeanstalk-amzn-2016.09.1.x86_64-ecs-pv-201704030647 (ami-059ea063).

When ever a new instance gets booted up the process fails. I find that the ecs daemon isn't running with the following ecs-init logs:

```
2017-05-19T09:29:55Z [INFO] pre-start
2017-05-19T09:29:56Z [INFO] start
2017-05-19T09:29:56Z [INFO] No existing agent container to remove.
2017-05-19T09:29:56Z [INFO] Starting Amazon EC2 Container Service Agent
2017-05-19T09:30:37Z [INFO] Agent exited with code 0
2017-05-19T09:30:37Z [INFO] Network error connecting to docker, backing off for '1.14777941s', error: dial unix /var/run/docker.sock: connect: no such file or directory
2017-05-19T09:30:40Z [INFO] post-stop
2017-05-19T09:30:40Z [INFO] Cleaning up the credentials endpoint setup for Amazon EC2 Container Service Agent
```
Here is what I find in the ecs-agent log:

```
2017-05-19T09:29:57Z [INFO] Starting Agent: Amazon ECS Agent - v1.14.1 (467c3d7)
2017-05-19T09:29:57Z [INFO] Loading configuration
2017-05-19T09:29:57Z [INFO] Checkpointing is enabled. Attempting to load state
2017-05-19T09:29:57Z [INFO] Loading state! module="statemanager"
2017-05-19T09:29:57Z [INFO] Event stream ContainerChange start listening...
2017-05-19T09:29:57Z [INFO] Detected Docker versions [1.17 1.18 1.19 1.20 1.21 1.22 1.23]
2017-05-19T09:29:57Z [INFO] Registering Instance with ECS
2017-05-19T09:29:58Z [INFO] Registered! module="api client"
2017-05-19T09:29:58Z [INFO] Registration completed successfully. I am running as 'arn:aws:ecs:eu-west-1:xxx:container-instance/79e9988c-8c80-4559-9d65-05807e5e63e0' in cluster 'awseb-xxx-b243eyymgn'
2017-05-19T09:29:58Z [INFO] Saving state! module="statemanager"
2017-05-19T09:29:58Z [INFO] Beginning Polling for updates
[...]

PR #142, opened 2017-12-27
Handle wrapped error for docker connect retry

jahkeup:
### Summary
Handles scenarios where the docker socket isn't yet available.

### Implementation details
I updated the condition in which the ping is retried.

### Testing
Tests were modified to reflect the wrapped error and I tested this with a bad path to be sure.

Fixed docker connection retry handling

### Licensing

### DIFF
--- a/ecs-init/docker/dependencies.go
+++ b/ecs-init/docker/dependencies.go
@@ -75,7 +75,7 @@ func newDockerClient(dockerClientFactory dockerClientFactory, pingBackoff backof
 			break
 		}
 		backoffDuration := pingBackoff.Duration()
-		log.Infof("Network error connecting to docker, backing off for '%v', error: %v", backoffDuration, err)
+		log.Infof("Error connecting to docker, backing off for %s, error: %s", backoffDuration, err)
 		time.Sleep(backoffDuration)
 	}
 	return &_dockerclient{
--- a/ecs-init/docker/dependencies.go
+++ b/ecs-init/docker/dependencies.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tdomhan:
Thanks a lot! So this would cover the case where the socket file does not yet exist and therefore solve issue https://github.com/aws/amazon-ecs-init/issues/92, is that correct?

ecs-init/docker/dependencies.go
```diff
@@ -131,15 +132,18 @@ func (s *_standardFS) ReadFile(filename string) ([]byte, error) {
 }
 
 func isNetworkError(err error) bool {
-	_, ok := err.(*net.OpError)
-	return ok
+	wrapped, isWrapped := err.(*url.Error)
+	if isWrapped {
```
aaithal: was this always a bug or did something change with the returned error from the library which resulted in this change?

jahkeup: It appears to me that we were not handling this properly. The Golang stdlib interactions should have been returning the wrapped error variant now handled appropriately. I didn't see anything in the upstream code (in `vendor/`) that suggests that this would have been unwrapped or returned differently. Our test case had a bad assumption (with a hardcoded `netErorr`) and we seemingly built up against that codified assumption.

jahkeup:
@tdomhan yes this should fix #92, `ecs-init` will now retry when connecting to-be-available socket with this PR.

ecs-init/docker/dependencies.go
aaithal: Thanks for that explanation. I wanted to verify that this change doesn't introduce different behavior with different versions of the daemon.

tdomhan:
So the retry logic seems to work fine now, however in total the backoff will only wait for 18s in total. Is there any way to configure this and increase the backoff time?

PR #147, opened 2018-02-05
Update to 1.17.0

jhaynes:
### Summary

Cache Agent 1.17.0

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -152,6 +152,8 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Mon Feb 05 2018 Justin Haynes <jushay@amazon.com> - 1.17.0-1
```
nmeyerhans: I believe the version in the RPM changelog entries should match the version that we're actually releasing. So here I think it should be 1.17.0-2. Actually I think it should be 1.17.0-2.amzn1, but I don't think we need that here.

PR #148, opened 2018-02-20
Cache agent 1.17.1.

jhaynes:
### Summary

Cache Agent 1.17.1

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -34,7 +34,7 @@ const (
 	ProcFS = "/proc"
 
 	// AgentFilename is the filename, including version number, of the agent to be downloaded.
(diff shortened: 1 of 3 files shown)


ecs-init/version/version.go
```diff
@@ -18,7 +18,7 @@
 package version
 
 // Version is the version of the ecs-init
-var Version string = "unknown"
+var Version string = "1.17.1"
```
jahkeup: I think we aren't changing this var anymore right? This is set at build time now (always was IIUC).

jhaynes: It was. I changed it out of habit and the update version script (called during build) sets it. Updated the PR to remove it.

PR #151, opened 2018-02-26
Add fluentd to default available logging drivers

amake:
### Summary
Add `fluentd` to the list of logging drivers available by default.

I have been unable to find a way to use fluentd as my logging driver without making a custom AMI. It would be great if it was usable by default.

Enhancement - Add `fluentd` to the list of logging drivers available by default

### Licensing


NightKhaos:
The following UserData snippet worked for me on the ecs-optimized AMI. User provided options in `/etc/ecs/ecs.config` SHOULD override the defaults. If they are not can you please provide more details?

    #!/bin/bash
    echo 'ECS_AVAILABLE_LOGGING_DRIVERS="json-file","syslog","awslogs","fluentd","none"]' >> /etc/ecs/ecs.config

amake:
Yes, that does indeed work. However in our case we are using multicontainers on Elastic Beanstalk and the only way to provide `ecs.config` requires us to create custom AMIs. This means extra work keeping up with changes to the base AMI just to add the config. We would love to be able to just use the base AMI as-is.

sparrc (approved):
looks good to me, have confirmed that the fluentd log driver is supported by our minimum docker api version in the agent:
see https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/logging_drivers.go#L38 and https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent.go#L732

PR #154, opened 2018-03-13
Enforce 8 character git short sha

jahkeup:
Enforce 8 character Git short SHAs.
### Licensing


jhaynes:
Can you update the commit message to include a description about why we need this change?

PR #159, opened 2018-03-29
Move to dep

petderek:
### Summary

Move from godep to dep

### Implementation details

something like:
```bash
dep init
rm -rf ./Godeps ./_vendor* 
make test
```

### Testing

New tests cover the changes: 
`make test`

### Licensing


Makefile
```diff
@@ -78,7 +78,7 @@ ubuntu-trusty:
 	cd BUILDROOT && debuild $(shell [ "$(DEB_SIGN)" -ne "0" ] || echo "-uc -us")
 
 get-deps:
-	go get github.com/tools/godep
+	go get github.com/x/lint/golint
```
richardpen: Can you also update the header of this file to 2018.

ecs-init/cache/dependencies_mock_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
richardpen: Are these changed because you updated the `gomock`?

petderek: Yes, and `make test` automatically updates the mocks in this project.

sharanyad: Is that good or bad? wondering what happens if there is error in updating mocks.

petderek: It means I have to fix them, which is annoying. :smile: 

@adnxn were just talking about solutions to this, but its probably something we wont spend a lot of time on here.

petderek:
Maybe the ref disappeared before travis could pull it?

I did something like this in the span of 60 seconds:
```bash
git rebase -i dev
git push --force petderek depdepdep
git checkout dev
git merge depdepdep
git push
```

Issue #157, opened 2018-03-27
use s3 client instead of http client to download images


aaithal:
### Summary

use s3 client instead of http client to download images

### Description

The `Downloader` uses `http` client to download the ECS agent tarball. It should use s3 client instead:
 https://github.com/aws/amazon-ecs-init/blob/24c976b4e04a369e8f25f7c48f159d712623824f/ecs-init/cache/cache.go#L39

samuelkarp:
If we move to the S3 client, we'll need to make sure to use `credentials.AnonymousCredentials` so we don't require the instance to have S3 permissions.

```go
svc := s3.New(session.Must(session.NewSession(&aws.Config{
  Credentials: credentials.AnonymousCredentials,
})))
```

aaithal:
We've released a new [version of the ECS init](https://github.com/aws/amazon-ecs-init/releases/tag/v1.17.3-1), which should fix this issue. Closing this for now.

PR #160, opened 2018-03-30
use s3client instead of httpclient in downloader

fenxiong:
### Summary

This pull request is implement the change from http client to s3 client, described in this issue: https://github.com/aws/amazon-ecs-init/issues/157.

### Implementation details

Use the s3 client instead of http client to download, and make some necessary changes due to the different of downloading interface.

### Testing

Automatic tests are added/updated accordingly.
Manual testing is done in us-west-2 by the following steps:
(1) Clean the cache directory (currently /tmp/var/cache/ecs/)
(2) Run 'make' in project directory
(3) Run the pre-start step of the amazon-ecs-init binary that is produced in (1). May need sudo.
(4) See the output of (3), and also check the cache directory.

New tests cover the changes: 
yes

Enhancement - use s3client instead of httpclient in downloader

### Licensing


ecs-init/cache/cache.go
```diff
@@ -57,6 +59,17 @@ func NewDownloader() *Downloader {
 		return downloader
 	}
 	downloader.metadata = ec2metadata.New(sessionInstance)
+
+	log.Debugf("Setting region for s3 client to: %s", downloader.getBucketRegion())
+	s3sessionInstance, err := session.NewSession(&aws.Config{
+		Credentials: credentials.AnonymousCredentials,
+		Region: aws.String(downloader.getBucketRegion()),
+	})
+	if err != nil {
```
aaithal: This will lead to silently consuming the error, which is less than ideal. Can you modify the signature of `NewDownloader` method return `error` and `Downloader` types and return the error when not `nil`?

ecs-init/cache/cache.go
```diff
@@ -91,6 +104,16 @@ func (d *Downloader) getRegion() string {
...
 }
 
+// getBucketRegion returns a region that contains the agent's bucket
+func (d *Downloader) getBucketRegion() string {
+	region := d.getRegion()
+	if config.IsSupportedRegion(region) {
+		return region
+	}
+
+	return config.DefaultRegionName
```
aaithal: Can you add a log line here stating that the supported region could not be determined?

fenxiong:
ok, will split it now

ecs-init/cache/cache.go
```diff
@@ -134,48 +154,69 @@ func (d *Downloader) DownloadAgent() error {
...
+	tempFile, err := d.fs.TempFile(config.CacheDirectory(), "ecs-agent.tar")
 	if err != nil {
-		return nil, err
+		return "", err
 	}
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected response code %d", resp.StatusCode)
+	log.Debugf("Temp file %s", tempFile.Name())
+	defer func() {
+		if err != nil {
```
aaithal: `err` gets reused multiple times in this method. Conditionally removing a file based on that can be erroneous depending on where/when `err` gets written to. Instead of doing this in `defer`, it could be less confusing to explicitly do it when this method runs into errors.

aaithal:
@fenxiong  can you please add details about any end-to-end tests you've done as well?

fenxiong:
Added the testing steps in the pr description.

fenxiong:
@aaithal I've fixed the problems you mentioned above. But about the explicitly removing the file problem, for the md5 case I just found out that I need to remove the temp file no matter there's error or not, so do I still need to remove the defer? Currently I only remove the defer for getting the agent tarball.

ecs-init/cache/cache.go
```diff
@@ -91,6 +104,18 @@ func (d *Downloader) getRegion() string {
...
 }
 
+// getBucketRegion returns a region that contains the agent's bucket
+func (d *Downloader) getBucketRegion() string {
+	region := d.getRegion()
+	if config.IsSupportedRegion(region) {
+		return region
+	}
+
+	log.Debugf("Current region does not contain the agent's s3 bucket, using the default region(%s) for downloader.",
```
adnxn: nit, logging wording: `"Current region (%s) does not contain...", region, config.DefaultRegionName`

ecs-init/cache/cache.go
```diff
@@ -36,16 +39,15 @@ const (
 
 // Downloader is responsible for cache operations relating to downloading the agent
 type Downloader struct {
-	getter   httpGetter
+	s3downloader s3Downloader
```
richardpen: This isn't formatted with gofmt.

ecs-init/cache/cache.go
```diff
@@ -104,28 +129,25 @@ func (d *Downloader) DownloadAgent() error {
...
+	publishedTarballReader, err := d.fs.Open(tempFileName)
 	if err != nil {
 		return err
 	}
-	log.Debugf("Temp file %s", tempFile.Name())
 	defer func() {
 		if err != nil {
-			log.Debugf("Removing temp file %s", tempFile.Name())
-			d.fs.Remove(tempFile.Name())
+			log.Debugf("Removing temp file %s", tempFileName)
```
adnxn: someone correct me if im wrong, but the `err != nil` check in the defer statement can have it's err value set to something else in L150 by `d.fs.Copy` before the defer statement actually executes. unless - is the intention to keep the tempfile around if the `d.fs.Copy` call fails? 

do we even need that `err != nil` check in the defer block?

ecs-init/cache/cache.go
```diff
@@ -54,10 +56,21 @@ func NewDownloader() *Downloader {
 	sessionInstance, err := session.NewSession()
 	if err != nil {
 		downloader.region = config.DefaultRegionName
```
richardpen: Do we still need this line, what's the behavior if an error was returned?

ecs-init/cache/cache.go
fenxiong: The intention is to remove any temp file if there's any error that fails the whole process. But I do find a mistake that I should move this defer func before its previous error check so that error from fs.Open will also result in deleting temp file.. I will update it
I think err!=nil check is needed so that we don't remove the temp file if we succeed (otherwise we will try to remove a non-existing file because it has been renamed to ecs-agent.tar)

ecs-init/cache/cache.go
```diff
@@ -134,48 +159,64 @@ func (d *Downloader) DownloadAgent() error {
...
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected response code %d", resp.StatusCode)
+	log.Debugf("Temp file %s", tempFile.Name())
+
+	_, err = d.s3downloader.Download(tempFile, &s3.GetObjectInput{
+		Bucket: aws.String(config.AgentRemoteBucketName),
+		Key: aws.String(config.AgentRemoteTarballKey()),
+	})
+	if err != nil {
+		tempFile.Close()
```
richardpen: Can you make this a `defer` after the create?

ecs-init/config/common.go
```diff
@@ -98,16 +107,19 @@ func AgentTarball() string {
 	return CacheDirectory() + "/ecs-agent.tar"
 }
 
-// AgentRemoteTarball is the remote location of the Agent image, used for populating the cache. This is retrieved
-// by region and the agent filename is appended.
-func AgentRemoteTarball(region string) string {
-	baseURI := getBaseLocationForRegion(region)
-	return baseURI + AgentFilename
+// AgentRemoteBucket returns the bucket that stores the Agent image
+func AgentRemoteBucket() string {
```
richardpen: What's the intention of this method, the `AgentRemoteBucketName` is already a global variable?

ecs-init/cache/cache.go
petderek: What do you think about doing a file existence check instead of an error check?

Your function here does the correct thing, but I'd rather not explicitly depend on scoping magic if we can avoid it.

ecs-init/cache/cache.go
fenxiong: I'm not sure why this line was needed before, so just leave it here. I will remove it for now.

ecs-init/cache/cache.go
fenxiong: yeah that's probably easier to understand. I will update it

ecs-init/cache/dependencies.go
```diff
@@ -23,16 +23,15 @@ package cache
...
-	"time"
+
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
-// httpGetter captures the only method used from the net/http package
-type httpGetter interface {
-	Get(url string) (resp *http.Response, err error)
+type s3Downloader interface {
```
aaithal: minor: please add documentation for this interface `// s3Downloader ...`

ecs-init/cache/cache.go
```diff
@@ -134,48 +156,67 @@ func (d *Downloader) DownloadAgent() error {
...
 func (d *Downloader) getPublishedMd5Sum() (string, error) {
-	region := d.getRegion()
-	agentRemoteTarballMD5 := config.AgentRemoteTarballMD5(region)
-	log.Debugf("Downloading published md5sum from %s", agentRemoteTarballMD5)
-	resp, err := d.getter.Get(agentRemoteTarballMD5)
+	tempFile, err := d.fs.TempFile(config.CacheDirectory(), "ecs-agent.tar.md5")
 	if err != nil {
 		return "", err
 	}
+	log.Debugf("Temp file %s", tempFile.Name())
```
aaithal: minor: `"Created temporary file for md5sum: %s"`

ecs-init/cache/cache.go
aaithal: This is still not addressed

fenxiong: I added it back due to another comment.. i will remove it now

ecs-init/cache/cache.go
```diff
@@ -134,48 +156,64 @@ func (d *Downloader) DownloadAgent() error {
...
-	resp, err := d.getter.Get(agentRemoteTarball)
+func (d *Downloader) getPublishedTarball() (string, error) {
+	tempFile, err := d.fs.TempFile(config.CacheDirectory(), "ecs-agent.tar")
 	if err != nil {
-		return nil, err
+		return "", err
 	}
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected response code %d", resp.StatusCode)
+	log.Debugf("Temp file %s", tempFile.Name())
```
aaithal: minor: `"Created temporary file for agent tarball: %s"`

ecs-init/cache/cache.go
```diff
@@ -36,28 +39,37 @@ const (
...
+		fs: &standardFS{},
 	}
 
 	// If metadata cannot be initialized the region string is populated with the default value to prevent future
 	// calls to retrieve the region from metadata
 	sessionInstance, err := session.NewSession()
 	if err != nil {
-		downloader.region = config.DefaultRegionName
-		return downloader
+		return nil, err
```
richardpen: Please correct me if I read it wrong. The sessionInstance is only used to get the region, while if it failed it should use the default region. But with this change, it's going to fail the download process?

ecs-init/cache/cache.go
adnxn: going over this again, seems like we can't defer for this specific call. if we go down this path with `err != nil` we need to close & remove the tempfile, and the deferred call to tempFile.Close() will try to close after the `d.fs.Remove` call.

ecs-init/cache/cache.go
fenxiong: it seems like so, but so is the original code... do you think we should ignore the error and just use the default region

aaithal: @richardpen is correct. that's a behavior change. let's maintain the same behavior as before with this PR. For now, logging the error and stating that default region is being used is good enough, while not failing the download itself.

fenxiong: i was wrong above.. now i understand. will update the code

ecs-init/cache/cache.go
```diff
@@ -36,28 +39,41 @@ const (
...
-		return downloader
+	} else {
+		// metadata is only used for retrieving the user's region
+		downloader.metadata = ec2metadata.New(sessionInstance)
 	}
-	downloader.metadata = ec2metadata.New(sessionInstance)
-	return downloader
+
+	log.Debugf("Setting region for s3 client to: %s", downloader.getBucketRegion())
+	s3sessionInstance, err := session.NewSession(&aws.Config{
```
aaithal: Why do we need 2 invocations of the `session.NewSession()` call? Seems like we should be fine with creating a new session just once?

fenxiong: I think we actually need two session instances because the session used for downloading from s3 needs a region to initialize, and that region is retrieved from the metadata, which needs a session instance... I don't find a method to set a region for session/download client

aaithal: I see. A minor nit here: can you rename `s3sessionInstance` to `s3Session`?

fenxiong: ok

PR #161, opened 2018-03-30
Update to Agent 1.17.3

jhaynes:
### Summary
Cache Agent 1.17.3

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -152,6 +152,8 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Fri Mar 30 2018 Justin Haynes <jushay@amazon.com> - 1.17.3-1
+- Cache Agent vesion 1.17.3
```
aaithal: I think https://github.com/aws/amazon-ecs-init/pull/160 warrants a changelog as well (as an `Enhancement`).

jahkeup: Yep! I think we may want to adopt the `CHANGELOG.md` procedure that the agent follows and port that to package specific changelogs at release time.

jhaynes: Agreed. For now, I'll add `- Use s3client instead of httpclient when downloading` to the spec file.

jahkeup: Added an issue to track this thought: https://github.com/aws/amazon-ecs-init/issues/162

PR #165, opened 2018-04-30
Bundle ECS Agent Tarball

jahkeup:
### Summary

This updates the specfile to include a bundled copy of the ECS Agent in the package itself. 

### Implementation details

The ECS Agent is added as a Source to the specfile and dropped into the `ecs-init` load directory. When the package is installed, the appropriate version of the 

### Testing

The resulting package was installed on an EC2 instance.

Bundle the respective ECS Agent in the `ecs-init` package.

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -95,6 +165,7 @@ touch $RPM_BUILD_ROOT/%{cache_dir}/state
 %config(noreplace) %ghost %{conf_dir}/ecs.config
 %config(noreplace) %ghost %{conf_dir}/ecs.config.json
 %ghost %{cache_dir}/ecs-agent.tar
```
samuelkarp: Should this remain `%ghost`?  It looks like we're now placing a symlink at this path.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -33,24 +34,93 @@ Requires:       iptables
...
+Provides:       bundled(golang(github.com/opencontainers/runc/libcontainer/user))
+Provides:       bundled(golang(github.com/pkg/errors))
 Provides:       bundled(golang(github.com/Sirupsen/logrus))
-Provides:       bundled(golang(github.com/cihub/seelog))
+Provides:       bundled(golang(golang.org/x/net/context))
+Provides:       bundled(golang(golang.org/x/net/context/ctxhttp))
+Provides:       bundled(golang(golang.org/x/sys/unix))
+Provides:       bundled(golang(golang.org/x/sys/windows))
 
+%global bundled_agent_version %{version}
```
samuelkarp: I don't see where `%{version}` is defined.  Should we read this from `ecs-init/VERSION`?

It might also be worthwhile passing `%{bundled_agent_version}` into the the build script (maybe as an argument or environment variable in the `%build` section) so we don't need to [maintain the constant](https://github.com/aws/amazon-ecs-init/blob/77120e36977944f8f78a3861b623aa53c3f11bcb/ecs-init/config/common.go#L36-L37) separately in the source anymore

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: Yes, this allows the package to own this file without needing to distribute the file (which is symlinked on install).

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: Version is [defined by the spec file](https://rpm-packaging-guide.github.io/#rpm-macros), we're tying this to the `ecs-init` Version until it diverges.

We could pass on the parameter into the build scripts, however externalizing that option would help the package configure `ecs-init`, either way it's something we can certainly consider in another iteration.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -33,24 +34,93 @@ Requires:       iptables
...
-Provides:       bundled(golang(github.com/docker/docker/pkg/units))
-Provides:       bundled(golang(github.com/docker/docker/vendor/src/code.google.com/p/go/src/pkg/archive/tar))
+Provides:       bundled(golang(github.com/docker/docker/pkg/term))
+Provides:       bundled(golang(github.com/docker/docker/pkg/term/windows))
+Provides:       bundled(golang(github.com/docker/go-connections/nat))
+Provides:       bundled(golang(github.com/docker/go-units))
 Provides:       bundled(golang(github.com/fsouza/go-dockerclient))
+Provides:       bundled(golang(github.com/go-ini/ini))
 Provides:       bundled(golang(github.com/golang/mock/gomock))
-Provides:       bundled(golang-logrus)
```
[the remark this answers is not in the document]
jahkeup: If I understand the packaging recommendations surrounding the use of bundled and de-bundled libraries, this would be `bundled(golang-logrus)` only if our package was consuming the named rpm for compilation, but we use our own vendored bundled copy.

packaging/amazon-linux-ami/ecs-init.spec
jhaynes: I'm inclined to stick with this behavior rather than changing the way we invoke the build.

jahkeup:
There's some other consideration that's needed for package upgrades. Working on that now.

jahkeup:
Please take another look if you haven't already. The follow up code change accounts for the `ecs-init` package's need to assert a reload of the bundled image on package updates. I have tested this with a few upgrade and downgrade paths and feel confident in its functionality. Please let me know if you know of other avenues that should be tested.

- [X] Image loaded by pre-start after fresh install
- [X] Image reloaded by pre-start after package is upgraded
- [X] Image reloaded by pre-start when package is downgraded
- [X] Image NOT reloaded when the UpdateContainerAgent API call is used

samuelkarp:
> Image NOT reloaded when the UpdateContainerAgent API call is used

Can you clarify the intended behavior here?  Right now, ecs-init will reload the image every time an upgrade or downgrade happens through yum because the state file is reset.  Are you intending to change that behavior?

ecs-init/cache/cache.go
```diff
@@ -37,6 +37,21 @@ const (
 	orwPerm = 0700
 )
 
+type CacheStatus uint8
+
+const (
+	// StatusUncached indicates that there is not an already downloaded
+	// and cached agent that is suitable for loading.
+	StatusUncached CacheStatus = iota
```
samuelkarp: If you're serializing these constants (which you are), please don't use `iota`.  It should be explicit when these numbers change meaning rather than implicit.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -107,48 +178,54 @@ rm -rf $RPM_BUILD_ROOT
 # record whether or not our service was running when docker is upgraded
 ecs_status=$(/sbin/status ecs 2>/dev/null || :)
 if grep -qF "start/" <<< "${ecs_status}"; then
-	/sbin/stop ecs >/dev/null 2>&1 || :
-	if [ "$1" -ge 1 ]; then
-		# write semaphore if this package is still installed
-		touch %{running_semaphore} >/dev/null 2>&1 || :
-	fi
+        /sbin/stop ecs >/dev/null 2>&1 || :
```
samuelkarp: I understand changing tabs to spaces, but why 8 spaces?

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -107,48 +178,54 @@ rm -rf $RPM_BUILD_ROOT
...
-	rm %{running_semaphore} >/dev/null 2>&1 ||:
+        /sbin/start ecs >/dev/null 2>&1 || :
+        rm %{running_semaphore} >/dev/null 2>&1 ||:
 fi
 
+%post
+# Symlink the bundled ECS Agent at loadable path.
+ln -sf ecs-agent-v%{bundled_agent_version}.tar %{cache_dir}/ecs-agent.tar
+# Indicate reload from cached image is needed.
+echo 2 > %{cache_dir}/state
```
samuelkarp: Instead of writing `2` in `%post`, should we just install this file with the content as `2`?

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: There must have been two tabs here, this was from expanding the tabs. I can clean up the entire file including spacing like this to fix up the extra indent.

ecs-init/cache/cache.go
jahkeup: Sounds good.

jahkeup:
@samuelkarp, that is the intended behavior. The agent docker container image will be reloaded on any changes to the rpm package - upgrade or downgrade. If a user were to call the UpdateContainerAgent API to upgrade the agent and not make any `ecs-init` package changes then the signaled upgraded image will be used. If a user were to call UpdateContainerAgent and then upgrade the `ecs-init` package (or downgrade) then they would change the version of the agent loaded respective to the installed package.

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: I don't see why not! I'll update the packaging.

ecs-init/cache/cache.go
```diff
@@ -37,6 +37,21 @@ const (
 	orwPerm = 0700
 )
 
+type CacheStatus uint8
```
aaithal: lint: missing documentation

ecs-init/cache/cache.go
```diff
@@ -76,11 +91,34 @@ func NewDownloader() (*Downloader, error) {
 	return downloader, nil
 }
 
+func (d *Downloader) AgentCacheStatus() CacheStatus {
+	stateFile := config.CacheState()
+	cacheControlled := d.fileNotEmpty(stateFile) && d.fileNotEmpty(config.AgentTarball())
```
aaithal: can this be named just `uncached`?

ecs-init/cache/cache.go
```diff
@@ -76,11 +91,34 @@ func NewDownloader() (*Downloader, error) {
 	return downloader, nil
 }
 
+func (d *Downloader) AgentCacheStatus() CacheStatus {
```
aaithal: It'd be good to add documentation for this method as well. More for understanding and for reference, it'd also be really useful to add a blurb about which cache status is returned when. For example, it's not obvious to me when `StatusReloadNeeded` will be returned from this method (unless someone happens to be looking at the rpm spec file).

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -24,6 +24,7 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
 Source0:        sources.tgz
 Source1:        ecs.conf
+Source2:        https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-v%{bundled_agent_version}.tar
```
aaithal: Which entity is responsible for generating this object? Is that being done in of ECS agent's release/publish scripts?

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -33,24 +34,93 @@ Requires:       iptables
 Requires:       procps
 Requires:       dhclient
 
-Provides:       bundled(docker)
+# find ../../ecs-init/vendor -name \*.go -exec dirname {} \; | sort | uniq | sed 's,^.*ecs-init/vendor/,,; s/^/bundled(golang(/;s/$/))/;' | sed 's/^/Provides:\t/' | expand -
+Provides:       bundled(golang(github.com/aws/aws-sdk-go/aws))
```
aaithal: Again, for my own understanding, does this mean that every time something new is vendored for ecs-init package, something needs to be added here as well? Can you please add some documentation regarding that as well if that's the case?

ecs-init/cache/cache_test.go
```diff
@@ -99,6 +102,46 @@ func TestIsAgentCachedTrue(t *testing.T) {
...
+
+			mockFS.EXPECT().Stat(config.CacheState()).Return(mockFSInfo, nil)
+			mockFS.EXPECT().Stat(config.AgentTarball()).Return(mockFSInfo, nil)
+			mockFSInfo.EXPECT().Size().Return(int64(1)).Times(2)
+			mockFS.EXPECT().Open(config.CacheState()).Return(file, nil)
+
+			d := &Downloader{fs: mockFS}
+
+			actual := d.AgentCacheStatus()
+			if actual != testcase.expected {
```
aaithal: nit: you can just do `assert.Equals` here (and in other tests)?

ecs-init/cache/cache.go
jahkeup: This is more like `cached`, I called it `cacheControlled` in that the presence of both non-empty files makes this a reportable status.

ecs-init/cache/cache.go
jahkeup: Agreed. I'll add some docs.

ecs-init/cache/cache.go
jahkeup: I'll add some docs.

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: This is an RPM Source declaration, today we don't have a wrapper to fetch this artifact automatically. Standard tools like `spectool` can be used to read this and fetch the appropriate source.

packaging/amazon-linux-ami/ecs-init.spec
jahkeup: Yes this should be updated when dependencies are updated. I'll find a good place to document this requirement.

ecs-init/cache/cache_test.go
jahkeup: I could, however I was avoiding adding new dependencies.

ecs-init/cache/cache.go
```diff
@@ -76,11 +99,38 @@ func NewDownloader() (*Downloader, error) {
...
+	file, err := d.fs.Open(stateFile)
+	if err != nil {
+		return StatusUncached
+	}
+	var status CacheStatus
+	_, err = fmt.Fscanf(file, "%d", &status)
+	if err != nil {
+		return StatusUncached
+	}
+	return status
```
samuelkarp: It looks like you could have an invalid `CacheStatus` like `3` be returned here.  It would be better to force that to `StatusUncached` than ignore it.

ecs-init/cache/cache_test.go
```diff
@@ -99,6 +102,46 @@ func TestIsAgentCachedTrue(t *testing.T) {
...
+	}{
+		// Expected states:
+		{"0", StatusUncached},
+		{"1", StatusCached},
+		{"2", StatusReloadNeeded},
+		{"1\n", StatusCached},
+		// Invalid states:
+		{"spurious", StatusUncached},
+		{" ", StatusUncached},
+		{"256", StatusUncached},
```
samuelkarp: Can you add `{"3", StatusUncached}`?

ecs-init/cache/cache.go
jahkeup: Yes, you can have an invalid status returned here. This was intentionally left unhandled.

ecs-init/cache/cache_test.go
jahkeup: This was intentionally unhandled, the decisions on handling a CacheStatus is on the caller, not this func.

PR #166, opened 2018-05-01
mount Docker plugin files dir

sharanyad:
### Summary
Mount Docker plugin files' directories in agent

### Implementation details
Docker plugins socket, spec and json files are present in three different directories on the host. More details [here](https://docs.docker.com/engine/extend/plugin_api/#plugin-discovery). Mount these directories to agent to use it for volume plugins discovery. 

### Testing

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -83,7 +83,22 @@ const (
 	CapSysAdmin = "SYS_ADMIN"
 	// DefaultCgroupMountpoint is the default mount point for the cgroup subsystem
 	DefaultCgroupMountpoint = "/sys/fs/cgroup"
+	// pluginSocketFilesDir specifies the location of UNIX domain socket files of
+	// Docker plugins
+	pluginSocketFilesDir = "/run/docker/plugins"
+	// pluginSpecFilesEtcDir specifies one of the locations of spec or json files
+	// of Docker plugins
+	pluginSpecFilesEtcDir = "/etc/docker/plugins"
+	// pluginSpecFilesUsrDir specifies one of the locations of spec or json files
+	// of Docker plugins
+	pluginSpecFilesUsrDir = "/usr/lib/docker/plugins"
 )
+
+var pluginDirs = []string{
+	pluginSocketFilesDir,
+	pluginSpecFilesEtcDir,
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker.go
```diff
@@ -261,9 +276,18 @@ func (c *Client) getHostConfig() *godocker.HostConfig {
 		config.CacheDirectory() + ":" + config.CacheDirectory(),
 		config.CgroupMountpoint() + ":" + DefaultCgroupMountpoint,
 	}
+	binds = append(binds, getDockerPluginDirBinds()...)
 	return createHostConfig(binds)
 }
 
+func getDockerPluginDirBinds() []string {
```
richardpen: Should we mount these directories readonly?

ecs-init/docker/docker_test.go
```diff
@@ -237,8 +237,8 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
 
 	hostCfg := opts.HostConfig
 
-	if len(hostCfg.Binds) != 10 {
-		t.Errorf("Expected exactly 10 elements to be in Binds, but was %d", len(hostCfg.Binds))
+	if len(hostCfg.Binds) != 13 {
```
jahkeup: Is there a better way to express this expectation?

petderek: For a unit test? I think this is reasonable when combined with the assertion later on.

jahkeup: @petderek I'm okay with this value here in the unit test, but it is a magic number that you have to know about and reconsider when it causes your tests to fail - it isn't a well defined number with its logic explained. This is the second or third time I've seen this change and, without context, wondered why we should assert that it was 13 (other than watching the test fail, and thinking 'oh yeah!'). I was hoping we had a better way to establish what this ought to be with the volume mount information at hand. It would be great to call out the conditions that it would change if fixing the test isn't practical.

sharanyad: Made this magic number a constant and added a note about it.

jahkeup: Awesome, thanks for adding that!

PR #167, opened 2018-05-03
Download the agent based on partition

richardpen:
### Testing

make test
New tests cover the changes: 
yes
### Licensing

yes

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -16,6 +16,9 @@ package config
 import (
 	"os"
 	"strings"
+
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ecs-init/config/common_test.go
```diff
@@ -47,22 +48,38 @@ func TestDockerUnixSocketWithDockerHost(t *testing.T) {
...
+			err:         nil,
+		}, {
+			region:      "us-gov-west-1",
+			destination: "us-gov-west-1",
+		}, {
+			region:      "cn-north-1",
+			destination: "cn-north-1",
+		}, {
+			region:      "invalid",
+			destination: "",
```
jahkeup: I think we could also then use the err field to assert there should be an error for this testcase, or we could get rid of the field entirely.

ecs-init/config/common_test.go
```diff
@@ -47,22 +48,38 @@ func TestDockerUnixSocketWithDockerHost(t *testing.T) {
...
+		destination string
+		err         error
 	}{
-		{DefaultRegionName, true},
-		{"cn-north-1", true},
-		{"missing-region", false},
+		{
+			region:      "us-west-2",
+			destination: "us-east-1",
+			err:         nil,
```
jahkeup: `err` isn't needed here.

ecs-init/config/common.go
```diff
@@ -44,17 +47,25 @@ const (
 	DefaultRegionName = "us-east-1"
```
jahkeup: Is this unused now?

richardpen: It's still used as a default if for some unknown reason the region isn't mapped to any of the partitions.

ecs-init/config/common.go
```diff
@@ -44,17 +47,25 @@ const (
...
 )
 
-// supportedRegions stores a list of regions that contains the agent bucket
-var supportedRegions = []string{"cn-north-1", "us-gov-west-1", DefaultRegionName}
+var partitionBucketMap = map[string]string{
+	"aws":        "us-east-1",
```
samuelkarp: You can use [these constants](https://github.com/aws/aws-sdk-go/blob/9a2fe34af9644afba4a1a1406966e78eb0e985af/aws/endpoints/defaults.go#L11-L13) and [these constants](https://github.com/aws/aws-sdk-go/blob/9a2fe34af9644afba4a1a1406966e78eb0e985af/aws/endpoints/defaults.go#L16-L44)

```go
var partitionBucketMap = map[string]string{
	endpoints.AwsPartitionID:      endpoints.UsEast1RegionID,
	endpoints.AwsCnPartitionID:    endpoints.CnNorth1RegionID,
	endpoints.AwsUsGovPartitionID: endpoints.UsGovWest1RegionID,
}
```

ecs-init/config/common.go
samuelkarp: `endpoints.UsEast1RegionID`

PR #168, opened 2018-05-04
Update to Agent 1.17.4

haikuoliu:
### Summary

Cache Agent 1.17.4

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -152,6 +152,10 @@ if [ -e %{running_semaphore} ]; then
 fi
 
 %changelog
+* Fri May 04 2018 Haikuo Liu <haikuo@amazon.com> - 1.17.4-1
+- Cache Agent vesion 1.17.4
+- Download agent based on the partition
+- Mount Docker plugin files dir
```
richardpen: I don't think you need the last two entries.

jahkeup: This is something we have done in the past for clarity. There is no other changelog communicated.

PR #170, opened 2018-05-15
Add support for regional buckets

fenxiong:
### Summary

This pull request adds support for regional buckets.

### Implementation details

The s3 downloader has two bucket downloaders internally, one corresponds to the partition bucket, another corresponds to the regional bucket. the download will succeed if either downloading from either bucket succeeds.

### Testing

make test
Manual test is performed on the produced rpm.

New tests cover the changes: 
yes

### Licensing

### DIFF
--- a/ecs-init/cache/cache.go
+++ b/ecs-init/cache/cache.go
@@ -24,17 +24,16 @@ import (
 	"strings"
 
 	"github.com/aws/amazon-ecs-init/ecs-init/config"
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
+
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/s3"
-	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	log "github.com/cihub/seelog"
+	"github.com/pkg/errors"
 )
 
 const (
-	orwPerm = 0700
+	orwPerm              = 0700
+	regionalBucketFormat = "%s-%s"
 )
 
 // CacheStatus represents the status of the on-disk cache for agent
@@ -62,7 +61,7 @@ const (
 
 // Downloader is responsible for cache operations relating to downloading the agent
 type Downloader struct {
-	s3downloader s3Downloader
+	s3Downloader s3DownloaderAPI
 	fs           fileSystem
 	metadata     instanceMetadata
 	region       string
@@ -78,7 +77,7 @@ func NewDownloader() (*Downloader, error) {
 	// calls to retrieve the region from metadata
 	sessionInstance, err := session.NewSession()
 	if err != nil {
-		log.Debugf("Get error when initializing session instance: %v. Use default region: %s",
+		log.Debugf("Got error when initializing session for metadata client: %v. Use default region: %s",
 			err, config.DefaultRegionName)
 		downloader.region = config.DefaultRegionName
 	} else {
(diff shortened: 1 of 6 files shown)


ecs-init/cache/dependencies.go
```diff
@@ -26,15 +26,103 @@ import (
...
 
+type s3BucketDownloaderAPI interface {
+	download(fileName, cacheDir string, fs fileSystem) (string, error)
+}
+
+// s3BucketDownloader wraps a bucket together with a downloader that can download from it
+type s3BucketDownloader struct {
+	bucket string
+	region string
+	s3     s3API
```
adnxn: nit: maybe change the name `s3` to `client`

ecs-init/cache/dependencies.go
```diff
@@ -26,15 +26,103 @@ import (
...
+
+	_, err = bd.s3.Download(file, &s3.GetObjectInput{
+		Bucket: aws.String(bd.bucket),
+		Key:    aws.String(fileName),
+	})
+
+	return file.Name(), err
+}
+
+type s3DownloaderAPI interface {
```
adnxn: hm do we really need to put this behind an interface?

ecs-init/cache/cache.go
```diff
@@ -222,53 +240,39 @@ func (d *Downloader) DownloadAgent() error {
...
-	}()
-	defer tempFile.Close()
 
-	_, err = d.s3downloader.Download(tempFile, &s3.GetObjectInput{
-		Bucket: aws.String(config.AgentRemoteBucketName),
-		Key:    aws.String(config.AgentRemoteTarballMD5Key()),
-	})
+	tempMd5File, err := d.fs.Open(tempMd5FileName)
 	if err != nil {
+		log.Debugf("Failed to open temporary md5 file: %v", err)
```
aaithal: I think all of these errors will be printed in the `main()`, here: https://github.com/aws/amazon-ecs-init/blob/8076adae902a81cb3669e1790883d6dc726c5473/ecs-init/ecs-init.go#L120. There's no need for this additional `Debubg`  log. However, it is important to capture the context of the error. Can you please wrap the original error with something like `errors.Wrapf(err, "failed to open temporary md5 file")`?

ecs-init/cache/cache.go
```diff
@@ -222,53 +240,39 @@ func (d *Downloader) DownloadAgent() error {
...
 	}
+	defer func() { // clean up temp file
+		log.Debugf("Removing temp file %s", tempMd5FileName)
+		d.fs.Remove(tempMd5FileName)
+	}()
 
-	body, err := d.fs.ReadAll(tempFile)
+	body, err := d.fs.ReadAll(tempMd5File)
 	if err != nil {
+		log.Debugf("Failed to read from temporary md5 file: %v", err)
```
aaithal: Same as above. Can you please wrap the original error with something like `errors.Wrapf(err, "failed to read from temporary md5 file")`?

ecs-init/cache/cache.go
```diff
@@ -222,53 +240,39 @@ func (d *Downloader) DownloadAgent() error {
...
-	_, err = d.s3downloader.Download(tempFile, &s3.GetObjectInput{
-		Bucket: aws.String(config.AgentRemoteBucketName),
-		Key:    aws.String(config.AgentRemoteTarballKey()),
-	})
+	tempAgentFileName, err := d.s3Downloader.downloadFile(config.AgentRemoteTarballKey())
 	if err != nil {
-		tempFile.Close()
-		log.Debugf("Removing temp file %s", tempFile.Name())
-		d.fs.Remove(tempFile.Name())
+		log.Debugf("Failed to download published tarball: %v", err)
```
aaithal: Same as above. Can you please wrap the original error with something like `errors.Wrapf(err, "failed to download published tarball")`?

ecs-init/cache/dependencies.go
```diff
@@ -26,15 +26,103 @@ import (
...
+	s3     s3API
+}
+
+func newS3BucketDownloader(region, bucketName string) (*s3BucketDownloader, error) {
+	session, err := session.NewSession(&aws.Config{
+		Credentials: credentials.AnonymousCredentials,
+		Region:      aws.String(region),
+	})
+	if err != nil {
+		log.Debugf("Got error when initializing downloader in region %s", region)
```
aaithal: Same as above. Can you please wrap the original error with something like `errors.Wrapf(err, "failed to initialize downloader in region %s", region)`?

ecs-init/cache/dependencies.go
fenxiong: I think this one is needed so that i can write the test

ecs-init/cache/cache.go
```diff
@@ -78,24 +77,44 @@ func NewDownloader() (*Downloader, error) {
...
+		cacheDir:          config.CacheDirectory(),
+		fs:                downloader.fs,
+	}
+
+	partitionBucketRegion := downloader.getPartitionBucketRegion()
+	partitionBucket := config.AgentPartitionBucketName
+	partitionBucketDownloader, err := newS3BucketDownloader(partitionBucketRegion, partitionBucket)
 	if err != nil {
-		return nil, err
+		log.Debugf("Failed to initialize partition bucket downloader: %v", err)
```
richardpen: This should be a `Warnf`

ecs-init/cache/cache.go
```diff
@@ -78,24 +77,44 @@ func NewDownloader() (*Downloader, error) {
...
+	} else {
+		s3Downloader.addBucketDownloader(partitionBucketDownloader)
+	}
+
+	region := downloader.getRegion()
+	regionalBucket := fmt.Sprintf(regionalBucketFormat, partitionBucket, region)
+	regionalBucketDownloader, err := newS3BucketDownloader(region, regionalBucket)
+	if err != nil {
+		log.Debugf("Failed to initialize regional bucket downloader: %v", err)
```
richardpen: Same as above.

ecs-init/cache/cache.go
```diff
@@ -78,24 +77,44 @@ func NewDownloader() (*Downloader, error) {
...
+	} else {
+		s3Downloader.addBucketDownloader(regionalBucketDownloader)
+	}
+
+	if len(s3Downloader.bucketDownloaders) == 0 {
+		log.Debug("Failed to initialize s3 downloader for either partition bucket or regional bucket. Downloader initialization fails.")
```
richardpen: This should be an `Error`

ecs-init/cache/dependencies.go
```diff
@@ -26,15 +26,98 @@ import (
...
+	fs                fileSystem
+	cacheDir          string
+}
+
+func (d *s3Downloader) addBucketDownloader(bucketDownloader *s3BucketDownloader) {
+	d.bucketDownloaders = append(d.bucketDownloaders, bucketDownloader)
+}
+
+func (d *s3Downloader) downloadFile(fileName string) (string, error) {
+	for _, bucketDownloader := range d.bucketDownloaders {
```
richardpen: I don't like the idea to implicitly retry with different downloaders. A better approach could be explicitly call the download function of the different downloader. This can be done in a separate pr if you want, so that this is not blocked.
```
downloadWithPartitionBucket()
downloadWithRegionalBucket()
```

fenxiong: i'm doing it implicitly due to other comments.. i guess we can discuss this offline and change it in another pr later

PR #171, opened 2018-05-31
Add ecs systemd unit for al2

jahkeup:
### Summary

This change wraps the existing `ecs-init` daemon (with minor semantic modifications) with a systemd unit to run under Amazon Linux 2.

### Implementation details

Minor modifications have been made to the entrypoints in the `ecs-init` binary to include the `stop` verb to align with systemd semantics surrounding service lifecycle. Though there is no difference from `post-stop` from upstart land, this allows the two to coexist in code if a change in handling is required at a later time.

The updated RPM spec accommodates both Amazon Linux and Amazon Linux 2 (thanks @ilianaw!) and includes a unit file suitable for running `ecs-init` on Amazon Linux 2 (with `systemctl start ecs`).

### Testing

The RPM was built both for Amazon Linux and Amazon Linux 2 and run in an ECS Cluster with some tasks to confirm no functional differences were present.

Add Amazon Linux 2 systemd unit to RPM spec

### Licensing

### DIFF
--- a/ecs-init/ecs-init.go
+++ b/ecs-init/ecs-init.go
@@ -31,6 +31,7 @@ const (
 	PRESTART = "pre-start"
 	START    = "start"
 	PRESTOP  = "pre-stop"
+	STOP     = "stop"
 	POSTSTOP = "post-stop"
 	RECACHE  = "reload-cache"
 )
@@ -91,7 +92,13 @@ func actions(engine *engine.Engine) map[string]action {
 			function:    engine.StartSupervised,
 			description: "Start the ECS Agent and wait for it to stop",
 		},
+		// This is a deprecated command for stopping the agent
+		// when using upstart jobs
 		PRESTOP: action{
+			function:    engine.PreStop,
+			description: "Stop the ECS Agent",
+		},
+		STOP: action{
 			function:    engine.PreStop,
 			description: "Stop the ECS Agent",
 		},
--- a/packaging/amazon-linux-ami/ecs-init.spec
+++ b/packaging/amazon-linux-ami/ecs-init.spec
@@ -12,25 +12,36 @@
 # License for the specific language governing permissions and
 # limitations under the License.
 
+%if 0%{?amzn} >= 2
+%bcond_without systemd # with
+%else
+%bcond_with systemd # without
+%global running_semaphore %{_localstatedir}/run/ecs-init.was-running
+%endif
+%global _cachedir %{_localstatedir}/cache
+%global bundled_agent_version %{version}
+
 Name:           ecs-init
 Version:        1.18.0
-Release:        1%{?dist}
-Group:          System Environment/Base
-Vendor:         Amazon.com
+Release:        2%{?dist}
 License:        Apache 2.0
 Summary:        Amazon Elastic Container Service initialization application
-BuildArch:      x86_64
-BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+ExclusiveArch:  x86_64
 
 Source0:        sources.tgz
 Source1:        ecs.conf
 Source2:        https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-v%{bundled_agent_version}.tar
+Source3:        ecs.service
 
 BuildRequires:  golang >= 1.7
-
-Requires:       docker >= 17.06.2ce
+%if %{with systemd}
+BuildRequires:  systemd
+Requires:       systemd
+%else
 Requires:       upstart
+%endif
 Requires:       iptables
+Requires:       docker >= 17.06.2ce
 Requires:       procps
 Requires:       dhclient
 
@@ -124,17 +135,6 @@ Provides:       bundled(golang(golang.org/x/net/context/ctxhttp))
 Provides:       bundled(golang(golang.org/x/sys/unix))
 Provides:       bundled(golang(golang.org/x/sys/windows))
 
-%global bundled_agent_version %{version}
-%global init_dir %{_sysconfdir}/init
-%global bin_dir %{_libexecdir}
-%global conf_dir %{_sysconfdir}/ecs
-%global cache_dir %{_localstatedir}/cache/ecs
-%global data_dir %{_sharedstatedir}/ecs/data
-%global dhclient_dir %{_sharedstatedir}/ecs/dhclient
-%global man_dir %{_mandir}/man1
-%global rpmstate_dir /var/run
-%global running_semaphore %{rpmstate_dir}/ecs-init.was-running
-
 %description
 ecs-init is a service which may be run to register an EC2 instance as an Amazon
 ECS Container Instance.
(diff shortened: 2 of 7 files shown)


packaging/amazon-linux-ami/ecs.service
```diff
@@ -0,0 +1,13 @@
...
+After=docker.service
+
+[Service]
+ExecStartPre=/usr/libexec/amazon-ecs-init pre-start
+ExecStart=/usr/libexec/amazon-ecs-init start
+ExecStop=/usr/libexec/amazon-ecs-init stop
+ExecStopPost=/usr/libexec/amazon-ecs-init post-stop
+
+[Install]
+WantedBy=multi-user.target
```
ilianaw: Please add a newline

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -12,25 +12,36 @@
...
 Version:        1.18.0
-Release:        1%{?dist}
-Group:          System Environment/Base
-Vendor:         Amazon.com
+Release:        2%{?dist}
 License:        Apache 2.0
 Summary:        Amazon Elastic Container Service initialization application
-BuildArch:      x86_64
-BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+ExclusiveArch:  x86_64
```
nmeyerhans: Why change from BuildArch to ExclusiveArch?

ecs-init/ecs-init.go
```diff
@@ -31,6 +31,7 @@ const (
 	PRESTART = "pre-start"
 	START    = "start"
 	PRESTOP  = "pre-stop"
+	STOP     = "stop"
```
nmeyerhans: You don't appear to have updated the manpage to reflect this change.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -144,230 +144,292 @@ ECS Container Instance.
...
-mkdir -p $RPM_BUILD_ROOT/%{man_dir}
-
-install %{SOURCE1} $RPM_BUILD_ROOT/%{init_dir}/ecs.conf
-install %{SOURCE2} $RPM_BUILD_ROOT/%{cache_dir}/ecs-agent-v%{bundled_agent_version}.tar
-install amazon-ecs-init $RPM_BUILD_ROOT/%{bin_dir}/amazon-ecs-init
-install scripts/amazon-ecs-init.1.gz $RPM_BUILD_ROOT/%{man_dir}/amazon-ecs-init.1.gz
-touch $RPM_BUILD_ROOT/%{conf_dir}/ecs.config
-touch $RPM_BUILD_ROOT/%{conf_dir}/ecs.config.json
+install -D amazon-ecs-init %{buildroot}%{_libexecdir}/amazon-ecs-init
+install -D scripts/amazon-ecs-init.1 %{buildroot}%{_mandir}/man1/amazon-ecs-init.1
```
nmeyerhans: We were explicitly compressing this before, but aren't here. Why?

packaging/amazon-linux-ami/ecs.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=ECS Agent
+Requires=docker.service
+After=docker.service
+
+[Service]
```
nmeyerhans: I wouldn't mind seeing `Type=simple` set explicitly here. Otherwise people will need to understand the rules around the different defaults in order to understand this unit's behavior.

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -12,25 +12,36 @@
...
+%global running_semaphore %{_localstatedir}/run/ecs-init.was-running
+%endif
+%global _cachedir %{_localstatedir}/cache
+%global bundled_agent_version %{version}
+
 Name:           ecs-init
 Version:        1.18.0
-Release:        1%{?dist}
-Group:          System Environment/Base
-Vendor:         Amazon.com
```
nmeyerhans: Why do we no longer set Group or Vendor?

packaging/amazon-linux-ami/ecs-init.spec
ilianaw: https://fedoraproject.org/wiki/Packaging:Guidelines#Arch-Specific_Runtime_and_Build-Time_Dependencies

packaging/amazon-linux-ami/ecs-init.spec
ilianaw: https://fedoraproject.org/wiki/Packaging:Guidelines#Manpages

>  When installing man pages, note that they should be installed uncompressed as the build system will compress them as needed.

packaging/amazon-linux-ami/ecs-init.spec
ilianaw: https://fedoraproject.org/wiki/Packaging:Guidelines#Tags_and_Sections

nmeyerhans: Why is that? Does that apply to all packages? It's not clear from that doc if there are fedora-specific reasons to avoid these tags or if they must not be present even in third-party packages.

ilianaw: AL2 inherits the packaging guidelines from Fedora / EPEL, except for the ones we are explicitly incompatible with (such as ones requiring changes in RPM 4.14) or otherwise simply don't like (none of these are documented yet).

The Group field, I believe, is not listed in yum repodata (deprecated in favor of yum's groups in comps.xml). The Vendor field gets overwritten by our build system anyway.

ecs-init/ecs-init.go
jahkeup: Indeed! I'll update and clarify how these targets should be used in the context of upstart and systemd.

packaging/amazon-linux-ami/ecs.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
samuelkarp: Can you add the license header to this file?

jahkeup:
[See rendered man page here](https://gist.github.com/jahkeup/63287c591e37e807d8e0d580f37423b3). Please leave comments in the PR and file.

jahkeup:
Trimmed down the commits. @samuelkarp PTAL at the man page updates specifically.

packaging/amazon-linux-ami/ecs.service
```diff
@@ -0,0 +1,28 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=ECS Agent
+Requires=docker.service
+After=docker.service
```
samuelkarp: We should probably have this run after `cloud-init.target` so that you can configure the ECS agent using user-data.

PR #175, opened 2018-06-13
Improve ecs systemd unit

jahkeup:
### Summary

This change ensures `ecs.service` start after `cloud-final.service`, the last cloud-init stage ([see docs][0]), for user data configuration of ecs-init. There are transitive `Requires` and `Wants` from other dependent units (ie: though `docker.service`) that achieve a desired ordering however including `After=cloud-final.service` ensures dependent units' ordering do not change `ecs.service`'s order requirement.

10s delay is added between `amazon-ecs-init` restart attempts. This gives some back off time to restart `ecs-init` (eg: more than 100ms).

[0]: https://cloudinit.readthedocs.io/en/0.7.9/topics/boot.html

### Implementation details

Appropriate systemd unit directives were added.

### Testing

System starts up with dependencies in the correct order.

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ecs.service
+++ b/packaging/amazon-linux-ami/ecs.service
@@ -16,6 +16,7 @@
 Description=ECS Agent
 Requires=docker.service
 After=docker.service
+After=cloud-final.service
 
 [Service]
 Type=simple
--- a/packaging/amazon-linux-ami/ecs.service
+++ b/packaging/amazon-linux-ami/ecs.service
@@ -20,6 +20,8 @@ After=cloud-final.service
 
 [Service]
 Type=simple
+Restart=on-failure
+RestartSec=10s
 ExecStartPre=/usr/libexec/amazon-ecs-init pre-start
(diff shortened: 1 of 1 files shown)


packaging/amazon-linux-ami/ecs.service
```diff
@@ -16,9 +16,12 @@
 Description=ECS Agent
 Requires=docker.service
 After=docker.service
+After=cloud-final.service
 
 [Service]
 Type=simple
+Restart=on-failure
+RestartSec=10s
```
jhaynes: Should we add a `StartLimitIntervalSec` or `StartLimitBurst`? Or just continue to restart forever? (restarting forever matches the behavior of the upstart managed service.

jahkeup: There are certainly more options we can start adding, but we really should have ecs-init itself start cooperating and allowing the init system to manage the process to have meaningful effect.

jhaynes: sgtm, :boat: :mortar_board:

PR #176, opened 2018-06-14
Allow cgroup mountpoint to be specified by env variable

jahkeup:
### Summary

This enables users to specify the environment variable `CGROUP_MOUNTPOINT` to direct `ecs-init` to use the provided path instead of the compiled default path.

### Testing

Unit test is added and the service was used on an AL2 instance where a task's cgroup is setup.

The environment variable is consulted before falling back to the system's compiled in default path.

### Licensing

### DIFF
--- a/ecs-init/config/common_test.go
+++ b/ecs-init/config/common_test.go
@@ -82,3 +82,36 @@ func TestGetAgentPartitionBucketRegion(t *testing.T) {
 			})
 	}
 }
+
+func TestCgroupMountpoint(t *testing.T) {
+	originalEnv := os.Getenv(cgroupMountpointEnv)
+	defer os.Setenv(cgroupMountpointEnv, originalEnv)
+
+	testcases := []struct {
+		name     string
+		env      string
+		expected string
+	}{
+		{
+			name:     "system default",
+			env:      "",
+			expected: cgroupMountpoint,
+		},
+		{
+			name:     "from environment",
+			env:      "/other/mountpoint/cgroup",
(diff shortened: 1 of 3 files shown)


ecs-init/config/common_test.go
```diff
@@ -82,3 +82,36 @@ func TestGetAgentPartitionBucketRegion(t *testing.T) {
...
+		env      string
+		expected string
+	}{
+		{
+			name: "system default",
+			env:  "", expected: cgroupMountpoint,
+		},
+		{
+			name: "from environment",
+			env:  "/other/mountpoint/cgroup", expected: "/other/mountpoint/cgroup",
```
jhaynes: Nit: Unless it is a go convention, I'd rather `expected:` be on a separate line.

ecs-init/config/common_test.go
```diff
@@ -82,3 +82,36 @@ func TestGetAgentPartitionBucketRegion(t *testing.T) {
 			})
 	}
 }
+
+const EnvCgroupMountpoint = "CGROUP_MOUNTPOINT"
```
aaithal: nit: this can be `envCgroupMountpoint`

jahkeup: Whoops, I meant for it to be using the const in common.go

ecs-init/config/common_test.go
jahkeup: Nope, I added the name and left the rest on the single line. Happy to add a new line for our eyes.

PR #177, opened 2018-06-15
Use /etc/ecs/ecs.config for cgroup path with unit default

jahkeup:
### Summary

This enables the use of the unit file and the existing /etc/ecs/ecs.config file to configure the init process as well. With this change, the cgroup mountpoint in and outside the spawned container will match and respect the same environment variable.

### Implementation details

The environment variable is reused from the ECS agent and loaded by the systemd unit when present. Defaults are provided in the unit. 

### Testing

Use `/etc/ecs/ecs.config` to align some settings

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -1,4 +1,4 @@
-// Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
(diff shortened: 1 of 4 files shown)


jahkeup:
This default environment variable would not be set when running from shell and could be confusing for customers trying to run `amazon-ecs-init` directly, any thoughts on how we could improve this?

ecs-init/config/common.go
```diff
@@ -47,8 +47,10 @@ const (
 	DefaultRegionName = endpoints.UsEast1RegionID
 
 	// cgroupMountpointEnv is the Environment Variable used to provide
-	// an alternative path to the cgroup mount on the host.
-	cgroupMountpointEnv = "CGROUP_MOUNTPOINT"
```
jhaynes: While I doubt the number is very high, there is potential for customers to be using this old environment variable. I think we need to maintain `CGROUP_MOUNTPONT` as an option too.

jahkeup:
Offline conversation with @jhaynes and @aaithal concluded that we'll continue with the build tag approach for supported platforms. We can revisit this (again) in the future when a larger and clearer use-case arises. Please take a look at #178 for the build tag solution.

PR #179, opened 2018-06-20
Add the attempted download URL to ecs-init error logging

kavoor:
### Summary
Added the attempted download URL to the Error logs when the download fails to allow engineers to more quickly determine the nature of the AccessDenied error.
Modified previous download failure message logging level from Debug to Error.

### Implementation details
Added line in the downloadFile() method containing the attempted download URL log statement. 
Changed function in prior failure message from Debugf() to Errorf().

### Testing

Ran 'make test'.

New tests cover the changes:  No new tests. 

Add the attempted download URL to ecs-init error logging.

### Licensing

### DIFF
--- a/ecs-init/cache/dependencies.go
+++ b/ecs-init/cache/dependencies.go
@@ -109,7 +109,7 @@ func (d *s3Downloader) downloadFile(fileName string) (string, error) {
 				fileName, bucketDownloader.bucket, bucketDownloader.region)
 			return fileName, nil
 		} else {
-			log.Debugf("Download file %s from bucket %s in region %s failed with error: %v",
+			log.Errorf("Download file %s from bucket %s in region %s failed with error: %v",
 				fileName, bucketDownloader.bucket, bucketDownloader.region, err)
 		}
 	}


ecs-init/cache/dependencies.go
```diff
@@ -109,8 +109,10 @@ func (d *s3Downloader) downloadFile(fileName string) (string, error) {
 				fileName, bucketDownloader.bucket, bucketDownloader.region)
 			return fileName, nil
 		} else {
-			log.Debugf("Download file %s from bucket %s in region %s failed with error: %v",
+			log.Errorf("Download file %s from bucket %s in region %s failed with error: %v",
 				fileName, bucketDownloader.bucket, bucketDownloader.region, err)
+			log.Errorf("Download from https://s3-%s.amazonaws.com/%s/%s failed.",
```
aaithal: i'd get rid of `https://` bit here because init doesn't download it from the URL. Instead, it uses the s3 client. instead, i'd prefer sticking in the region in the error log at line 112

jahkeup: This could also be formatted as `s3://%s/%s` for the bucket and key attempted - making it a usable s3 uri given a partition and along with the respective region which is included here that might conveys enough to debug with.

aaithal: S3 has a weird URL scheme, which can vary depending on region, private link support etc. So, i'd prefer to not log anything that resembles a URL. We can include additional information if anything is missing when things go awry to help us debug the issue, but just not in a URL format. 

https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

PR #181, opened 2018-07-05
Update systemd unit description and add url

jahkeup:
### Summary

Update systemd unit description and add documentation url.

### Implementation details

Directives were updated in the systemd units.

### Testing

I dropped the changed unit into `/etc/systemd/system` and ran the following:

```yaml
# systemctl status ecs.service
● ecs.service - Amazon Elastic Container Service - container agent
   Loaded: loaded (/etc/systemd/system/ecs.service; disabled; vendor preset: enabled)
   Active: inactive (dead)
     Docs: https://aws.amazon.com/documentation/ecs/
```

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ecs.service
+++ b/packaging/amazon-linux-ami/ecs.service
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 [Unit]
-Description=ECS Agent
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
 Requires=docker.service
 After=docker.service
 After=cloud-final.service
--- a/packaging/suse/amazon-ecs.service
+++ b/packaging/suse/amazon-ecs.service
@@ -1,5 +1,6 @@
 [Unit]
-Description=Amazon Elastic Container Service Initialization
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
 After=docker.service
 After=network.target
 Requires=docker.service


packaging/amazon-linux-ami/ecs.service
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 [Unit]
-Description=ECS Agent
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
```
samuelkarp: You could change this to:

```
Documentation=https://aws.amazon.com/documentation/ecs/ https://ecs.aws
```

packaging/suse/amazon-ecs.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
-Description=Amazon Elastic Container Service Initialization
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
```
samuelkarp: Same as above

packaging/amazon-linux-ami/ecs.service
jahkeup: That's a fun url! However, I would prefer to have links that direct users to the documentation pages rather than a product overview page. Is there a path on that domain that points to more specific documentation for the agent or init?

samuelkarp: Not that I know of, but that's why I left the documentation URL first.  `Documentation` supports a space-separated list of URLs.

jahkeup:
@jhaynes, it's been a while since you've had eyes on this, can you take a look and confirm this change looks good?

Issue #190, opened 2018-08-13
ecs-agent's json log file can get really big


jhaynes:
### Summary
While the log files in `/var/log/ecs` are rotated, the json file that lives at `/var/lib/docker/containers/<sha>/` does not rotate.

### Description
Starting the ecs-agent container differently can mitigate this issue via `docker run` flags.

jahkeup:
This was taken care of with the merge of #195 and will be included in the next release of ecs-init. Closing.

PR #192, opened 2018-08-28
Setting a LogConfig to prevent excessive disk usage of container

NightKhaos:
### Summary
This pull request adds a LogConfig to the HostConfig of the running ECS Agent Container in order to prevent excessive disk usage on hosts where there is a lot of audit log activity as noted in #190 

### Implementation details
Modified `docker_unspecified.go` to change the HostConfig to include a LogConfig that will limit disk usage of `stdout`/`stderr` log for container (using the `json-file` driver) to approximately 2GiB in the worst case.

Used the `max-file` and `max-size` options for `json-file` to achieve this. See [this page](https://docs.docker.com/config/containers/logging/json-file/).

Also made some minor fixes to fix where `x/lint/golint` came from and deal with the situation where compiling on a case insensitive filesystem (VERSION file and version directory conflict).

### Testing
Compiled the agent with changes and uploaded to an ECS Optimized AMI, overriding the binary.
Started and Stopped the Agent using `ecs stop; ecs start` and ensured that the new running agent container had the configured log configuration options.

Restricted `json-file` log usage to 2GiB.

### Licensing


haikuoliu:
Hi @NightKhaos 

Thanks so much for your contribution, I changed the base branch from `master` to `dev` as we develop on `dev` branch and merge `dev` to `master` when we release.

yunhee-l:
Why was VERSION file renamed? is this needed for reducing disk usage by logs?

NightKhaos:
Hi @yunhee-l, I explained this in my blurb:

>  ...deal with the situation where compiling on a case insensitive filesystem (VERSION file and version directory conflict)

On some operating systems, most notably Mac OS X, the default filesystem configuration is case insensitive. The existence of a directory and a file named the same thing on this operating system prevented me from completing the build as the VERSION file would never actually be created when cloning (as the VERSION file would conflict with the version directory).

Probably something you want to address at some point, even if you decide not to accept it as part of this commit.

ecs-init/docker/docker_unspecified.go
```diff
@@ -40,7 +40,16 @@ func createHostConfig(binds []string) *godocker.HostConfig {
 		dhclientLibDir+":"+dhclientLibDir+readOnly,
 		dhclientExecutableDir+":"+dhclientExecutableDir+readOnly)
 
+	// Bit ugly adding the log rotation configuration here inline, consider
+	// revising to make this easier to read.
 	return &godocker.HostConfig{
+		LogConfig:   godocker.LogConfig{
+			Type: "json-file",
+			Config: map[string]string{
+				"max-size": "256m",
```
jahkeup: This number seems more than reasonable to me; what is the reasoning behind this choice?

This arrangement (`8*256m`) will still result in some slightly large log files totalling 2GB, and AFAIK this is duplicated log output that the agent writes to `/var/log/ecs`. There may be some tools out there using `docker logs` to obtain these logs so they definitely could be taking advantage of the slack - however if they are consuming near-realtime then retaining 2GB of duplicated logs may be unnecessary. 

It may be prudent to start with smaller values and allow the knobs to be tuned here for use cases where a larger retention policy is needed. What do you think about that approach?

scripts/gobuild.sh
```diff
@@ -20,7 +20,7 @@ export GOPATH="${TOPWD}/ecs-init/:${BUILDDIR}"
 export SRCPATH="${BUILDDIR}/src/github.com/aws/amazon-ecs-init"
 
 if [ -d "${TOPWD}/.git" ]; then
-    version=$(cat "${TOPWD}/ecs-init/VERSION")
+    version=$(cat "${TOPWD}/ecs-init/ECSVERSION")
```
jahkeup: I think I've heard peeps of discontent around this before, thanks for fixing this!

ecs-init/docker/docker_unspecified.go
```diff
@@ -40,7 +40,16 @@ func createHostConfig(binds []string) *godocker.HostConfig {
 		dhclientLibDir+":"+dhclientLibDir+readOnly,
 		dhclientExecutableDir+":"+dhclientExecutableDir+readOnly)
 
+	// Bit ugly adding the log rotation configuration here inline, consider
```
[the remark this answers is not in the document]
NightKhaos: Refactored it to use the a config object and for readability in latest commit.

ecs-init/docker/docker_unspecified.go
NightKhaos: Seems reasonable. Have changed the starting values to 16m, 4 files and put them in a configuration item in ecs-init/config/common.go

NightKhaos:
@jahkeup Hi Jake, thanks for your feedback, I have changed the code to use a config object, and cleaned up a little bit for usability. Sorry for the messy commits, still learning Go at the moment. I stand by my suggestion to squash if you merge this.

ecs-init/config/common.go
```diff
@@ -138,12 +138,18 @@ func DockerUnixSocket() (string, bool) {
...
 }
 
 // CgroupMountpoint returns the cgroup mountpoint for the system
 func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
+
+// AgentStdoutStderrLoggingMaxSizeMaxFile returns a tuple of the Maximum log file size and maximum number of log files
+// to be rotated, e.g. 256m, 8
+func AgentStdoutStderrLoggingMaxSize() (string, string) {
```
jahkeup: This is headed down the right path, but in order for customers to customize this value they would need to reconfigure this by editing the file and recompiling. There are a few examples in this package that check with the environment variables to determine if the user would like a different setting to be returned. We'll want to give some thought to these variables though, these variables are somewhat unique as they are specific to `ecs-init` itself and we don't have many of them. Let me think a bit on naming and usage.

I prefer that we don't return two different, though closely related, strings from this function. It requires that we remember the order and meaning of each string at the call site. We can avoid the ambiguity by having a function that returns each discrete value (max size and files, so something like `func DockerLogMaxFiles()` and `func DockerLogMaxSize()`) *or* have a function that returns the `godocker.LogConfig{}` struct fully configured.

My vote would be for the multiple function approach, but would be open to consider another if you have suggestions.

NightKhaos: If we're going for use configurable, the customer MAY want to change their log driver from json-file. Therefore it makes sense to return a godocker.LogConfig{} struct and decide on some environment variables to configure that. I feel deciding on the environment variables is out of scope for this PR, thus I have left that as a TODO item to decide on later.

jahkeup:
@NightKhaos we'll take this change through #195 so we get the environment variables and consensus on supporting the tunables there. :+1: Thanks again for your contributions!

Issue #131, opened 2017-11-17
testing with suse, ubuntu tags


adnxn:
Today we aren't exercising test paths for non amazon linux operating systems.
Running `go test -short -v -cover -tags suse ./...` and `go test -short -v -cover -tags ubuntu ./...` results in test failures for `TestStartAgentNoEnvFile` and `TestStartAgentEnvFile`.

We should fix these and also add the tagged test runs to our CI.

ellenthsu:
closing this item as it's not an item the team will be working on in the near future

PR #198, opened 2018-09-28
config, docker: config certs dir path

adnxn:
### Summary

add host SSL cert directory path for AL and AL2

### Implementation details
added expected paths to platform specific config files behind build flags.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -150,7 +150,7 @@ func CgroupMountpoint() string {
 
 // HostCertsDirPath() returns the CA store path on the host
 func HostCertsDirPath() string {
-	if _, err := os.Stat(hostCertsDirPath); os.IsNotExist(err) {
+	if _, err := os.Stat(hostCertsDirPath); err != nil {
 		return ""
 	}
 	return hostCertsDirPath
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -148,8 +148,10 @@ func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
 
-// HostCertsDirPath() returns the CA store path on the host (specifically on
-// Amazon Linux platform)
+// HostCertsDirPath() returns the CA store path on the host
 func HostCertsDirPath() string {
+	if _, err := os.Stat(hostCertsDirPath); os.IsNotExist(err) {
+		return ""
+	}
 	return hostCertsDirPath
 }
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -147,3 +147,9 @@ func DockerUnixSocket() (string, bool) {
 func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
+
+// HostCertsDirPath() returns the CA store path on the host (specifically on
+// Amazon Linux platform)
+func HostCertsDirPath() string {
+	return hostCertsDirPath
+}
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -221,8 +221,8 @@ func (c *Client) getContainerConfig() *godocker.Config {
 	}
 
 	// for al, al2 add host ssl cert directory envvar if available
-	if config.HostCertsDirPath() != "" {
-		envVariables["SSL_CERT_DIR"] = config.HostCertsDirPath()
+	if certDir := config.HostCertsDirPath(); certDir != "" {
+		envVariables["SSL_CERT_DIR"] = certDir
 	}
 
 	// merge in platform-specific environment variables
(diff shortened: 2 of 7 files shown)


ecs-init/config/common.go
```diff
@@ -147,3 +147,9 @@ func DockerUnixSocket() (string, bool) {
 func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
+
+// HostCertsDirPath() returns the CA store path on the host (specifically on
+// Amazon Linux platform)
+func HostCertsDirPath() string {
+	return hostCertsDirPath
```
jahkeup: This should verify the directory exists even though in the context of Amazon Linux it is fairly benign and seemingly unneeded. Other platforms that aren't suse or ubuntu will be impacted by this change and would otherwise blindly set the certs dir path regardless of its presence.

ecs-init/docker/docker_test_unspecified.go
```diff
@@ -0,0 +1,23 @@
+// +build !suse !ubuntu
```
fenxiong: this should be // +build !suse,!ubuntu ?

ecs-init/config/common.go
adnxn: > Other platforms that aren't suse or ubuntu will be impacted by this change and would otherwise blindly set the certs dir path regardless of its presence.

fair. if path doesn't exist, will default to bundle.

ecs-init/docker/docker.go
```diff
@@ -220,6 +220,11 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		"ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST": "true",
 	}
 
+	// for al, al2 add host ssl cert directory envvar if available
+	if config.HostCertsDirPath() != "" {
```
jahkeup: nit: save the stat calls, `if certdir := config.HostCertsDirPath(); certdir != "" { }`

ecs-init/docker/docker.go
```diff
@@ -276,6 +281,13 @@ func (c *Client) getHostConfig() *godocker.HostConfig {
 		config.CacheDirectory() + ":" + config.CacheDirectory(),
 		config.CgroupMountpoint() + ":" + DefaultCgroupMountpoint,
 	}
+
+	// for al, al2 add host ssl cert directory mounts
+	if config.HostCertsDirPath() != "" {
```
jahkeup: nit: save the Stats(), `if certdir := config.HostCertsDirPath(); certdir != "" { }`

ecs-init/config/common.go
```diff
@@ -147,3 +147,12 @@ func DockerUnixSocket() (string, bool) {
 func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
+
+// HostCertsDirPath() returns the CA store path on the host (specifically on
+// Amazon Linux platform)
```
jahkeup: nit: This isn't specific to Amazon Linux, the configured value may be, I think this comment could simply trim off the parenthetical and remain perfectly clear.

jahkeup:
Looks like there may be a lingering test issue, can you take a look?

https://travis-ci.org/aws/amazon-ecs-init/builds/434802933#L587-L591

ecs-init/docker/docker.go
adnxn: oh right, that changed now. good catch thanks.

adnxn:
@jahkeup: right we still haven't fixed platform tagged testing, also just found related issue #131 

removing testing dependencies with build tags.

ecs-init/docker/docker_test.go
```diff
@@ -244,6 +242,14 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
 
 	hostCfg := opts.HostConfig
 
+	// for hosts that do not have cert directories explicity mounted, ignore
+	// host cert directory configuration.
+	// TODO (adnxn): ideally, these should be behind build flags.
+	// https://github.com/aws/amazon-ecs-init/issues/131
+	if certDir := config.HostCertsDirPath(); certDir == "" {
+		expectedAgentBinds = 13
```
jahkeup: nit: how about `-1`?

ecs-init/config/common.go
```diff
@@ -147,3 +147,11 @@ func DockerUnixSocket() (string, bool) {
 func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
+
+// HostCertsDirPath() returns the CA store path on the host
+func HostCertsDirPath() string {
+	if _, err := os.Stat(hostCertsDirPath); os.IsNotExist(err) {
```
samuelkarp: Any reason not to just use `err != nil` here?

adnxn: just chatted with @jahkeup about this, you're right we can just capture all errors.

PR #199, opened 2018-10-09
config, docker: add host /etc/pki dir as mount

adnxn:
### Summary

mounting root cert store director to agent container so cert store can be accessed by agent binary, additionally we're configuring the specific SSL_CERT_DIR to the paths expected on AL1 and AL2.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -155,3 +155,11 @@ func HostCertsDirPath() string {
 	}
 	return hostCertsDirPath
 }
+
+// HostPKIDirPath() returns the CA store path on the host
+func HostPKIDirPath() string {
+	if _, err := os.Stat(hostPKIDirPath); err != nil {
+		return ""
+	}
+	return hostPKIDirPath
+}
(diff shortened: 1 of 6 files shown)


yunhee-l:
Can you add details to summary and implementation details please.

adnxn:
> Can you add details to summary and implementation details please.

@yunhee-l: added

ecs-init/config/common.go
```diff
@@ -155,3 +155,11 @@ func HostCertsDirPath() string {
 	}
 	return hostCertsDirPath
 }
+
+// HostPKIDirPath() returns the CA store path on the host
+func HostPKIDirPath() string {
+	if _, err := os.Stat(hostPKIDirPath); err != nil {
```
samuelkarp: It looks like `hostPKIDirPath` is missing from `ecs-init/config/config_suse_ubuntu.go`, so compilation using those tags will fail.  Can you add a definition and then check here for `hostPKIDirPath != ""` as well?

adnxn: sigh, fixing.

PR #206, opened 2018-10-25
Initial GPU support

sharanyad:
### Summary
This PR integrates init with NVML APIs to obtain GPU information on the instance

### Implementation details
In the PR:
- `engine.go`:
In pre-start, GPU Manager is called for setting up and obtaining the gpu information on the instance, which is later bind mounted into agent 
- `nvidia_gpu_manager.go`: 
Functions that call NVML APIs to obtain GPU information on the instance and write it to disk for agent 
- Updated dep and THIRD-PARTY file
- updated mockgen to rename all `_mock_test.go` to `_mocks.go`

Not in the PR, but will add later:
- retries for NVML API failures
- Obtain Nvidia Docker Version in the instance

### Testing
unit tests pass

Manual testing on a p3 instance:
```
# cat /var/lib/ecs/gpu/nvidia-gpu-info.json | python -mjson.tool
{
    "DriverVersion": "396.44",
    "GPUIDs": [
        "GPU-962128ef-df85-bbed-2ae8-a3886dbb3135",
        "GPU-61806bb9-bd8e-5c38-4c6c-85df571ec13f",
        "GPU-d48b9dda-fec7-fd95-f7f6-da76d5c78884",
        "GPU-a4d06855-655f-6792-d785-23a18b02c6df"
    ],
    "NvidiaDockerVersion": ""
}
```

Adds support for GPU information detection

### Licensing


samuelkarp:
> updated mockgen to rename all _mock_test.go to _mocks.go

Is there a particular reason you wanted to do this?  Without the `_test.go` suffix, the mocks will actually be compiled into the binary instead of just being in the test binary.

ecs-init/gpu/nvidia_gpu_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	GPUIds              []string
+}
+
+const (
+	// NvidiaGPUDeviceFilePattern is the pattern of GPU device files on the instance
+	NvidiaGPUDeviceFilePattern = "/dev/nvidia*"
+	// GPUInfoDirPath is the directory where gpus and driver info are saved
+	GPUInfoDirPath = "/var/lib/ecs/gpu"
+	// NvidiaGPUInfoFilePath is the file path where gpus and driver info are saved
+	NvidiaGPUInfoFilePath = "/var/lib/ecs/gpu/nvidia-gpu-info.json"
```
haikuoliu: This can be `NvidiaGPUInfoFilePath = GPUInfoDirPath + /nvidia-gpu-info.json`

sharanyad:
> Is there a particular reason you wanted to do this? 

yes, go test files are not visible outside a package. so we cannot call methods(such as `NewMockSomething`) inside a mock file outside of that package.

aaithal:
> yes, go test files are not visible outside a package. so we cannot call methods(such as `NewMockSomething`) inside a mock file outside of that package.

That's strictly not true. In this [example](https://github.com/aws/amazon-ecs-agent/blob/3d3685542da83b8935c365366fc27486289b392b/agent/tcs/client/client_test.go#L136), the mock websocket connection method `mock_wsconn. NewMockWebsocketConn` is being invoked from a different package (`tcs/client`).

ecs-init/gpu/nvidia_gpu_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	DetectGPUDevices() error
+	SaveGPUState() error
+}
+
+// NvidiaGPUManager is used as a wrapper for NVML APIs and implements GPUManager
+// interface
+type NvidiaGPUManager struct {
+	DriverVersion       string
+	NvidiaDockerVersion string
+	GPUIds              []string
```
haikuoliu: This should be GPUIDs (tho it seems a bit wired). Same for other places.

ecs-init/gpu/nvidia_gpu_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "nvidia gpu manager: detecting GPU devices failed")
+	}
+	if matches == nil {
+		return ErrNoGPUDeviceFound
+	}
+	return nil
+}
+
+var MatchFilePattern = FilePatternMatch
```
haikuoliu: Why `MatchFilePattern` cannot be `matchFilePattern`? seems like we don't use it outside of this package.

ecs-init/gpu/nvidia_gpu_manager.go
```diff
@@ -0,0 +1,218 @@
...
+// Shutdown is for shutting down nvidia's nvml library
+func (n *NvidiaGPUManager) Shutdown() error {
+	err := ShutdownNVML()
+	if err != nil {
+		return errors.Wrapf(err, "nvidia gpu manager: error shutting down nvidia nvml")
+	}
+	return nil
+}
+
+var ShutdownNVML = ShutdownNVMLib
```
haikuoliu: Seems we don't have a consistent pattern for the naming, if would be better if we follow the patter something like below, that way we will be able to distinguish them by name:
```
var NVMLShutdown = Shutdown
func Shutdown() ...

var NVMLGetDriverVersion = GetDriverVersion
func GetDriverVersion()...
```

sharanyad:
> You don't need to modify that for all mock files, I guess is what I'm trying to say 

make test generates the mock files, hence changed mockgen which resulted in existing files changes too.

ecs-init/gpu/nvidia_gpu_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	return nvml.GetDriverVersion()
+}
+
+// GetGPUDeviceIDs is for getting the GPU device UUIDs
+func (n *NvidiaGPUManager) GetGPUDeviceIDs() ([]string, error) {
+	count, err := NvmlGetDeviceCount()
+	if err != nil {
+		return nil, errors.Wrapf(err, "nvidia gpu manager: error getting GPU device count for UUID detection")
+	}
+	var gpuIds []string
```
adnxn: nit: `gpuIDs`

PR #211, opened 2018-11-09
Consume custom instance configuration for GPU support

sharanyad:
### Summary
Custom GPU AMI configuration `ECS_ENABLE_GPU_SUPPORT` to be set at `/usr/lib/ecs/ecs.config` 
This will be further used for getting GPU information in the instance 
Note: This can be overridden by user config at `/etc/ecs/ecs.config`

### Testing

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -203,3 +203,13 @@ func AgentDockerLogDriverConfiguration() godocker.LogConfig {
 		},
 	}
 }
+
+// InstanceConfigDirectory returns the location on disk for custom instance configuration
+func InstanceConfigDirectory() string {
+	return directoryPrefix + "/usr/lib/ecs"
+}
+
+// InstanceConfigFile returns the location of a file of custom environment variables
+func InstanceConfigFile() string {
+	return InstanceConfigDirectory() + "/ecs.config"
+}
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -63,6 +63,8 @@ const (
 	// be used to override the default value used of
 	// dockerJSONLogMaxFiles for managed containers.
 	dockerJSONLogMaxFilesEnvVar = "ECS_INIT_DOCKER_LOG_FILE_NUM"
+	// GPUSupportEnvVar indicates that the AMI has support for GPU
+	GPUSupportEnvVar = "ECS_ENABLE_GPU_SUPPORT"
 )
 
 var partitionBucketMap = map[string]string{
@@ -206,7 +208,7 @@ func AgentDockerLogDriverConfiguration() godocker.LogConfig {
 
 // InstanceConfigDirectory returns the location on disk for custom instance configuration
 func InstanceConfigDirectory() string {
-	return directoryPrefix + "/usr/lib/ecs"
+	return directoryPrefix + "/var/lib/ecs"
 }
 
 // InstanceConfigFile returns the location of a file of custom environment variables
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -230,8 +230,13 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		envVariables[envKey] = envValue
 	}
 
+	// merge in instance-specific environment variables
+	for envKey, envValue := range c.LoadCustomInstanceEnvVars() {
+		envVariables[envKey] = envValue
+	}
+
 	// merge in user-supplied environment variables
-	for envKey, envValue := range c.loadEnvVariables() {
+	for envKey, envValue := range c.LoadEnvVariables() {
 		envVariables[envKey] = envValue
 	}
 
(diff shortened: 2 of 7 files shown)


jhaynes:
We'll need to update our packaging accordingly to include creation of this path and potentially the file if we want to put any defaults in it today.

jahkeup:
@sharanyad we'll need to ensure the packaging owns this directory so it exists. I have some feedback specifically regarding the path that I'll post with my review of the PR. Feel free to ping me offline too if you have any questions.

sharanyad:
> I have some feedback specifically regarding the path that I'll post with my review of the PR

Sure, I'll hold off on merging until then

ecs-init/config/common.go
```diff
@@ -203,3 +203,13 @@ func AgentDockerLogDriverConfiguration() godocker.LogConfig {
 		},
 	}
 }
+
+// InstanceConfigDirectory returns the location on disk for custom instance configuration
+func InstanceConfigDirectory() string {
+	return directoryPrefix + "/usr/lib/ecs"
```
jahkeup: I've given this some thought, chatted with @jhaynes a bit, and I think this should be `/var/lib/ecs`. 

Ideally, the packaging should be able to provide some "vendor" defaults that are included in the `ecs-init` package. Settings for baked or pre-configured AMIs would reside in `/var/lib/ecs`; user configured settings would reside in `/etc/ecs`. This particular change is targeting the case of pre-configuration, rather than vendor defaults, from what I can tell so I suggest we use `/var/lib/ecs`.

This layout would have us loading configuration in this order:

1st: `/usr/share/ecs`, the package's defaults
2nd: `/var/lib/ecs`, the system's baked in configuration (overloadable for use in the case of baked AMIs)
3rd: `/etc/ecs`, the user's general configuration

And though we aren't vending an `ecs-init` package today with configuration in `/usr/share/ecs`, we may want to consider doing so in order to provide the package a means to suggest appropriate defaults.

ecs-init/docker/docker.go
```diff
@@ -230,8 +230,13 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		envVariables[envKey] = envValue
 	}
 
+	// merge in instance-specific environment variables
+	for envKey, envValue := range c.LoadCustomInstanceEnvVars() {
+		envVariables[envKey] = envValue
```
jahkeup: Can we add some debug logging for when these values are overridden?

ecs-init/engine/dependencies.go
```diff
@@ -36,6 +36,8 @@ type dockerClient interface {
 	RemoveExistingAgentContainer() error
 	StartAgent() (int, error)
 	StopAgent() error
+	LoadEnvVariables() map[string]string
```
jahkeup: Rather than exposing both locations, I think it would be cleaner to have the logic of the merge fully encapsulated or moved into the config package.

ecs-init/docker/docker.go
```diff
@@ -246,10 +251,20 @@ func (c *Client) getContainerConfig() *godocker.Config {
 	}
 }
 
-func (c *Client) loadEnvVariables() map[string]string {
+// LoadEnvVariables gets user-supplied environment variables
+func (c *Client) LoadEnvVariables() map[string]string {
```
jahkeup: Why are these methods on the Client when these are products entirely of configuration? Is this here only for testing?

ecs-init/engine/engine.go
```diff
@@ -32,6 +32,7 @@ const (
 	terminalSuccessAgentExitCode = 0
 	terminalFailureAgentExitCode = 5
 	upgradeAgentExitCode         = 42
+	gpuSupportEnvVar             = "ECS_ENABLE_GPU_SUPPORT"
```
jahkeup: I think it would be cleaner to have the config package export env vars as consts rather than scattering them about the code base. Should we move these there?

ecs-init/docker/docker.go
sharanyad: this method is already present today. the method felt out of place, but I tried to use it as is and avoid too much refactoring work.

ecs-init/config/common.go
sharanyad: sounds good to me, I can change this to `/var/lib/ecs`

ecs-init/docker/docker.go
jahkeup: The current usage doesn't depend on the `*Client` here, this could be moved to the config package as a function there and have the Client hold or accept a reference to a `var environment map[string]string` for when it requires it. There shouldn't be need for consumers of the configured environment to have to know how to merge the variables.

I feel most strongly about having the merge logic hidden and managed centrally, I don't think we *must* have the implementation moved. It is certainly desirable though.

sharanyad: The existing design is that `Client` has filesystem embedded, which it uses to read the config files.
I agree this is not the best way to do this, but to change this, it will require moving the filesystem out to a different package like utils. 
I can add a `TODO` and do the refactoring work in a different PR. 

With regards to hiding the merging logic, `LoadEnvVariables()`  and `LoadCustomInstanceEnvVars()` can be combined into a single method that engine can use. Let me know what you think.

PR #214, opened 2018-11-15
enable cgo for nvidia code

sharanyad:
### Summary
address build error 
`go build github.com/aws/amazon-ecs-init/ecs-init/vendor/github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml: invalid flag in #cgo LDFLAGS: -Wl,--unresolved-symbols=ignore-in-object-files`

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/scripts/gobuild.sh
+++ b/scripts/gobuild.sh
@@ -37,14 +37,14 @@ mkdir -p "${SRCPATH}"
 ln -s "${TOPWD}/ecs-init" "${SRCPATH}"
 cd "${SRCPATH}/ecs-init"
 if [[ "$1" == "dev" ]]; then
-	go build -tags 'development' -ldflags "${VERSION_FLAG} ${GIT_HASH_FLAG} ${GIT_DIRTY_FLAG}" \
+	CGO_ENABLED=1 CGO_LDFLAGS_ALLOW='-Wl,--unresolved-symbols=ignore-in-object-files' go build -tags 'development' -ldflags "${VERSION_FLAG} ${GIT_HASH_FLAG} ${GIT_DIRTY_FLAG}" \
 	   -o "${TOPWD}/amazon-ecs-init"
 else
 	tags=""
 	if [[ "$1" != "" ]]; then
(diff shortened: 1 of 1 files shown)


petderek:
Did you check `make rpm`?

scripts/gobuild.sh
```diff
@@ -37,14 +37,14 @@ mkdir -p "${SRCPATH}"
 ln -s "${TOPWD}/ecs-init" "${SRCPATH}"
 cd "${SRCPATH}/ecs-init"
 if [[ "$1" == "dev" ]]; then
-	go build -tags 'development' -ldflags "${VERSION_FLAG} ${GIT_HASH_FLAG} ${GIT_DIRTY_FLAG}" \
+	CGO_ENABLED=1 CGO_LDFLAGS_ALLOW='-Wl,--unresolved-symbols=ignore-in-object-files' go build -tags 'development' -ldflags "${VERSION_FLAG} ${GIT_HASH_FLAG} ${GIT_DIRTY_FLAG}" \
```
petderek: nit: maybe put the flags into another variable since we repeat later

jahkeup: I couldn't find anything to clarify the `--unresolved-symbols=ignore-in-object-files` portion of the flag, can you give some background on why these linker options are required for building with CGO specifically for enabling nvidia code?

sharanyad: vendor package for nvml APIs uses the cgo ldflags here - https://github.com/aws/amazon-ecs-init/blob/84a9381fe5cba620a9e71b5a959c4854278eacac/ecs-init/vendor/github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml/bindings.go#L5
Looks like only certain ldflags are whitelisted by default (https://github.com/golang/go/issues/23749) and hence the build error as mentioned in the summary of this PR.

jahkeup: Thanks, that explains it :+1:

PR #215, opened 2018-11-15
Bump release to 3 for rebuild

jahkeup:
Bump NVR for a rebuild.


yunhee-l:
do you also need to update packaging/ubuntu-trusty/debian/changelog?

jahkeup:
Yes, we'll have that in an additional commit. This is the commit that was used to release the freshly released ecs-init package.

PR #217, opened 2018-11-27
mount gpu info dir only for GPU supported instances

sharanyad:
### Summary
same as title

### Implementation details
check if the instance config or env var `ECS_ENABLE_GPU_SUPPORT` is set to true and mount gpu info dir into agent container 

### Testing

N/A

### Licensing

### DIFF
--- a/ecs-init/docker/docker_test.go
+++ b/ecs-init/docker/docker_test.go
@@ -18,15 +18,17 @@ import (
 	"testing"
 
 	"github.com/aws/amazon-ecs-init/ecs-init/config"
+	"github.com/aws/amazon-ecs-init/ecs-init/gpu"
 	godocker "github.com/fsouza/go-dockerclient"
 	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
 )
 
 // expectedAgentBinds is the total number of agent host config binds.
 // Note: Change this value every time when a new bind mount is added to
 // agent for the tests to pass
 const (
-	expectedAgentBindsUnspecifiedPlatform = 16
+	expectedAgentBindsUnspecifiedPlatform = 15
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker_test.go
```diff
@@ -351,6 +352,45 @@ func TestStartAgentEnvFile(t *testing.T) {
...
+		expectedAgentBinds = expectedAgentBindsUnspecifiedPlatform
+	}()
+
+	mockFS := NewMockfileSystem(mockCtrl)
+	mockDocker := NewMockdockerclient(mockCtrl)
+
+	mockFS.EXPECT().ReadFile(config.InstanceConfigFile()).Return([]byte(envFile), nil).AnyTimes()
+	mockFS.EXPECT().ReadFile(config.AgentConfigFile()).Return(nil, errors.New("not found")).AnyTimes()
+	mockDocker.EXPECT().CreateContainer(gomock.Any()).Do(func(opts godocker.CreateContainerOptions) {
+		validateCommonCreateContainerOptions(opts, t)
```
haikuoliu: How did you verify that the GPU dir is actually binded here? Is `validateCommonCreateContainerOptions` already changed to verify the GPU bind mount?

sharanyad: in all these tests, we just verify the number of bind mounts rather than actual value. I'll check if we can verify GPU dir mount.

haikuoliu: I think it's validated in `validateCommonCreateContainerOptions`

https://github.com/haikuoliu/amazon-ecs-init/blob/master/ecs-init/docker/docker_test.go#L249

sharanyad: updated

PR #224, opened 2018-12-25
send GPU instance config only when GPUs are present

sharanyad:
### Summary
add GPU instance config variable `ECS_ENABLE_GPU_SUPPORT` to agent container only when GPU devices are present in the instance

### Testing
Manual testing:
Non GPU instance:
```
$ docker inspect ecs-agent | grep "ECS_ENABLE_GPU_SUPPORT"
[ec2-user@ip-172-31-18-88 ~]$
```
GPU instance:
```
$ docker inspect ecs-agent | grep "ECS_ENABLE_GPU_SUPPORT"
                "ECS_ENABLE_GPU_SUPPORT=true",
```

N/A

### Licensing

### DIFF
--- a/ecs-init/docker/docker_test.go
+++ b/ecs-init/docker/docker_test.go
@@ -482,6 +482,13 @@ func TestGetInstanceConfig(t *testing.T) {
 	defer mockCtrl.Finish()
 
 	envFile := "\nECS_ENABLE_GPU_SUPPORT=true\n"
+	defer func() {
+		MatchFilePatternForGPU = FilePatternMatchForGPU
+	}()
+	MatchFilePatternForGPU = func(pattern string) ([]string, error) {
+		// GPU device present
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker_test.go
```diff
@@ -482,6 +482,13 @@ func TestGetInstanceConfig(t *testing.T) {
 	defer mockCtrl.Finish()
 
 	envFile := "\nECS_ENABLE_GPU_SUPPORT=true\n"
+	defer func() {
+		MatchFilePatternForGPU = FilePatternMatchForGPU
```
petderek: NIt: to me, MatchFilePattern sounds like a regex pattern alone. I'd rather call it FilepathGlob or something like that.

sharanyad: we are infact doing a regex pattern check for `/dev/nvidia*`

PR #238, opened 2019-02-26
Add binding for agent container to be able to run iptables.

suneyz:
### Summary

Add binding /sbin, /usr/lib64, /lib64 to support iptable command for appmesh cni plugin

### Testing

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -84,6 +84,17 @@ const (
 	// pluginSpecFilesUsrDir specifies one of the locations of spec or json files
 	// of Docker plugins
 	pluginSpecFilesUsrDir = "/usr/lib/docker/plugins"
+	// iptablesExecutableDir specifies the location of the iptable
+	// executable on the host and in the Agent container
+	iptablesExecutableDir = "/sbin"
+	// iptablesLibDir specifies the location of shared libraries on the
+	// host and in the Agent container required for the execution of the iptables
+	// executable
+	iptablesLibDir = "/lib64"
+	// iptablesUsrLibDir specifies the location of shared libraries on the
+	// host and in the Agent container required for the execution of the iptables
+	// executable. Some OS like AL2 moved lib64 to /usr/lib64
+	iptablesUsrLibDir = "/usr/lib64"
 )
 
 var pluginDirs = []string{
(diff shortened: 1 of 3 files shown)


ecs-init/docker/docker.go
```diff
@@ -84,6 +84,17 @@ const (
 	// pluginSpecFilesUsrDir specifies one of the locations of spec or json files
 	// of Docker plugins
 	pluginSpecFilesUsrDir = "/usr/lib/docker/plugins"
+	// iptableExecutableDir specifies the location of the iptable
+	// executable on the  host and in the Agent container
```
yumex93: There seems an extra space between **the** and **host**.

ecs-init/docker/docker.go
```diff
@@ -84,6 +84,17 @@ const (
 	// pluginSpecFilesUsrDir specifies one of the locations of spec or json files
 	// of Docker plugins
 	pluginSpecFilesUsrDir = "/usr/lib/docker/plugins"
+	// iptableExecutableDir specifies the location of the iptable
```
aaithal: can you please %s/iptable/iptables/ everywhere?

suneyz: updated :)

suneyz:
> Can you please update the commit message with why the agent needs these bindings?

updated to include a bit more context of why we are adding this binding to support appmesh plugin

aaithal:
Thanks for updating the commit message. I think it should contain more information as to why you need to bind these things, especially the part about ECS agent container needing these. For example:

```
Bind additional dirs to agent to support aws-appmesh plugin

This is needed as the appmesh cni plugin needs to invoke `iptables` command. The 
ECS agent container doesn't have the necessary lib and binary paths to invoke 
this command. The additional volume mounts /sbin, /usr/lib64, /lib64 help with that.
```

suneyz:
> Thanks for updating the commit message. I think it should contain more information as to why you need to bind these things, especially the part about ECS agent container needing these. For example:
> 
> ```
> Bind additional dirs to agent to support aws-appmesh plugin
> 
> This is needed as the appmesh cni plugin needs to invoke `iptables` command. The 
> ECS agent container doesn't have the necessary lib and binary paths to invoke 
> this command. The additional volume mounts /sbin, /usr/lib64, /lib64 help with that.
> ```

Thanks for help! I will update to include more information you pointed out :)

PR #244, opened 2019-04-24
snap package

mikeroyal:
includes snapcraft.yaml file to build snap.

### Summary

Adds the snapcraft.yaml file to build a snap package.

### Testing

https://snapcraft.io/build
New tests cover the changes: 
yes
Snaps make installing your app easier and more secure.

Feature, Enhancement- makes it easier to install and update app across multiple Linux distros.

### Licensing

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,7 +1,8 @@
 name: amazon-ecs-init
 version: git
-summary: The Amazon Elastic Container Service RPM.
-description: 
+summary:  Elastic Container Service.
+description:|  
+The Amazon Elastic Container Service RPM.
   
 adopt-info: amazon-ecs-init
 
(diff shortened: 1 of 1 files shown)


petderek:
Its neat to see this! How does snap's strict confinement work?

Agent needs sys / net admin capabilities in order to manage docker containers within an ecs task. I don't know much about snap, but it looks like it prevents the sort of access agent requires.

mikeroyal:
Hi @petderek, we can fix that by changing the condiment to 'classic' instead. https://docs.snapcraft.io/snap-confinement/6233

sparrc:
We do have plans to support Ubuntu but probably will not be supporting that as a snap package.

Some of the current work towards that goal can be seen here: https://github.com/aws/amazon-ecs-init/pull/347

PR #248, opened 2019-05-25
service start: add backoff to service restarts

sparrc:
Because we hide the service's return code from systemd/sysV/upstart, we are not able to use the init system's builtin restart limiting.

With this change we can backoff restarting the service to avoid a situation where the service continuously restarts many times per second, resulting in numerous API calls to startup APIs with zero backoff, as each time the service starts it would be with a "fresh" API backoff counter.

### Licensing

### DIFF
--- a/ecs-init/engine/engine.go
+++ b/ecs-init/engine/engine.go
@@ -17,7 +17,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
+	"time"
 
+	"github.com/aws/amazon-ecs-init/ecs-init/backoff"
 	"github.com/aws/amazon-ecs-init/ecs-init/cache"
 	"github.com/aws/amazon-ecs-init/ecs-init/config"
 	"github.com/aws/amazon-ecs-init/ecs-init/docker"
@@ -33,6 +36,11 @@ const (
 	terminalSuccessAgentExitCode = 0
 	terminalFailureAgentExitCode = 5
 	upgradeAgentExitCode         = 42
+	serviceStartMinRetryTime     = time.Millisecond * 500
+	serviceStartMaxRetryTime     = time.Second * 15
+	serviceStartRetryJitter      = 0.10
+	serviceStartRetryMultiplier  = 2.0
+	serviceStartMaxRetries       = math.MaxInt64 // essentially retry forever
 )
 
 // Engine contains methods invoked when ecs-init is run
@@ -168,7 +176,9 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
 // StartSupervised starts the ECS Agent and ensures it stays running, except for terminal errors (indicated by an agent exit code of 5)
 func (e *Engine) StartSupervised() error {
 	agentExitCode := -1
-	for agentExitCode != terminalSuccessAgentExitCode && agentExitCode != terminalFailureAgentExitCode {
+	retryBackoff := backoff.NewBackoff(serviceStartMinRetryTime, serviceStartMaxRetryTime,
+		serviceStartRetryJitter, serviceStartRetryMultiplier, serviceStartMaxRetries)
+	for {
 		err := e.docker.RemoveExistingAgentContainer()
 		if err != nil {
 			return engineError("could not remove existing Agent container", err)
@@ -180,15 +190,24 @@ func (e *Engine) StartSupervised() error {
 			return engineError("could not start Agent", err)
 		}
 		log.Infof("Agent exited with code %d", agentExitCode)
-		if agentExitCode == upgradeAgentExitCode {
+
+		switch agentExitCode {
+		case upgradeAgentExitCode:
 			err = e.upgradeAgent()
 			if err != nil {
 				log.Error("could not upgrade agent", err)
+			} else {
+				// continuing here because a successful upgrade doesn't need to backoff retries
+				continue
 			}
+		case terminalFailureAgentExitCode:
+			return errors.New("agent exited with terminal exit code")
+		case terminalSuccessAgentExitCode:
+			return nil
 		}
-	}
-	if agentExitCode == terminalFailureAgentExitCode {
-		return errors.New("agent exited with terminal exit code")
+		d := retryBackoff.Duration()
+		log.Warnf("ECS Agent failed to start, retrying in %s", d)
+		time.Sleep(d)
 	}
 	return nil
 }


yunhee-l:
have we tested this change with backend? 1min+ delay seems a bit much, and I'm curious how this will impact P99 latency for task launches.

ecs-init/engine/engine.go
```diff
@@ -186,6 +214,7 @@ func (e *Engine) StartSupervised() error {
 				log.Error("could not upgrade agent", err)
 			}
 		}
+		retryCount++
```
fenxiong: i think this change adds 2 seconds delay for update container agent api (happy case).. can you skip adding retry count when e.upgradAgent() does not return error?

ecs-init/engine/engine.go
```diff
@@ -165,10 +167,36 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
...
+	return delay
+}
+
 // StartSupervised starts the ECS Agent and ensures it stays running, except for terminal errors (indicated by an agent exit code of 5)
 func (e *Engine) StartSupervised() error {
 	agentExitCode := -1
+	retryCount := 0
 	for agentExitCode != terminalSuccessAgentExitCode && agentExitCode != terminalFailureAgentExitCode {
+		if retryCount > 0 {
+			time.Sleep(retryDelay(retryCount))
```
fenxiong: can you log the delay being used here? i think it's better if customers (and even ourselves) know about how long we need to wait

ecs-init/engine/engine.go
```diff
@@ -165,10 +167,36 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
...
+//     Retry 2: 7.548s
+//     Retry 3: 14.776s
+//     Retry 4: 16.944s
+//     Retry 5: 34.592s
+//     Retry 6: 1m0.425s
+//     Retry 7: 1m0.456s
+//     Retry 8: 1m0.694s ...etc...
+func retryDelay(retryCount int) time.Duration {
+	minTime := 1000
+	maxTime := time.Second * 60
```
fenxiong: can you make 1000 and 60 * time.Second as consts in this package?

sparrc: I don't foresee these values getting used anywhere else so I'm not sure making constants is necessary here

ecs-init/engine/engine.go
sparrc: yes will do 👍

ecs-init/engine/engine.go
sparrc: good catch, will do 👍

ecs-init/engine/engine.go
```diff
@@ -168,7 +170,8 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
 // StartSupervised starts the ECS Agent and ensures it stays running, except for terminal errors (indicated by an agent exit code of 5)
 func (e *Engine) StartSupervised() error {
 	agentExitCode := -1
-	for agentExitCode != terminalSuccessAgentExitCode && agentExitCode != terminalFailureAgentExitCode {
+	backoff0 := backoff.NewBackoff(time.Millisecond*500, time.Second*15, 0.10, 2.0, 99999999)
```
jahkeup: The intent would be clearer with some constants to sweep away the magic numbers, consider something like:

```suggestion
	retryBackoff := backoff.NewBackoff(StartupRetryJitter, config.StartupRetrySeconds, DESCRIPTIVE_CONSTANT, DESCRIPTIVE_CONSTANT, DESCRIPTIVE_CONSTANT)
```

Allowing the user to passthrough some of these values could mean that in some cases system administrators could influence the unit's restart/retry policy, like on hosts running systemd.

ecs-init/engine/engine.go
```diff
@@ -180,15 +183,23 @@ func (e *Engine) StartSupervised() error {
...
+		case terminalFailureAgentExitCode:
+			return errors.New("agent exited with terminal exit code")
+		case terminalSuccessAgentExitCode:
+			return nil
 		}
-	}
-	if agentExitCode == terminalFailureAgentExitCode {
-		return errors.New("agent exited with terminal exit code")
+		d := backoff0.Duration()
+		log.Infof("Service failed to start, sleeping %s and retrying", d)
```
jahkeup: This shouldn't be an informational log line, its a warning that `ecs-init` wasn't able to start the agent but is going to retry again. Maybe use `log.Warnf` and a message like `Agent container startup failed, retrying startup in %s`.

ecs-init/engine/engine.go
sparrc: I think making this externally configurable is outside the scope of this change. That would also require updating documentation and messaging out the change. AFAIK there aren't any customers asking to configure the service startup retry interval of the ECS agent.

ecs-init/engine/engine.go
sparrc: sure, I'll change it to a warning 👍

ecs-init/engine/engine.go
sparrc: I have changed the random numbers to named local constants so it's more clear what each argument means.

ecs-init/engine/engine.go
```diff
@@ -168,7 +170,14 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
 // StartSupervised starts the ECS Agent and ensures it stays running, except for terminal errors (indicated by an agent exit code of 5)
 func (e *Engine) StartSupervised() error {
 	agentExitCode := -1
-	for agentExitCode != terminalSuccessAgentExitCode && agentExitCode != terminalFailureAgentExitCode {
+	const (
```
sharanyad: please move constants to the top of the file

ecs-init/engine/engine.go
```diff
@@ -168,7 +170,14 @@ func (e *Engine) load(image io.ReadCloser, err error) error {
...
+		minTime    = time.Millisecond * 500
+		maxTime    = time.Second * 15
+		jitter     = 0.10
+		multiplier = 2.0
+	)
+	backoff0 := backoff.NewBackoff(minTime, maxTime, jitter, multiplier, 99999999)
```
sharanyad: what is 99999999?

ecs-init/engine/engine.go
jahkeup: > I think making this externally configurable is outside the scope of this change. That would also require updating documentation and messaging out the change. AFAIK there aren't any customers asking to configure the service startup retry interval of the ECS agent.

I'm thinking about this more in the context of the packaging used on Amazon Linux where the init daemon can be entrusted to handle the retry cooperatively with this code path. This would allow for the systemd unit to specify a retry policy where the host would restart `ecs-init` when appropriate and it itself be the place for configuration.

The time to add configurables is when the code is added IMO. Not going to block the merge on it though.

A remaining nit: let's change `backoff0`, I proposed `retryBackoff`, what do you think about that?

sparrc: I'm not really a fan of the retry logic living in ecs-init rather than the system's init process, but that's what we have already and I think trying to break that apart is a larger and much riskier change.

..but I will change the variable name to retryBackoff 👍

ecs-init/engine/engine.go
sparrc: that's the max number of retries, it's not getting used in this case because we don't call "backoff.ShouldRetry" anyways

ecs-init/engine/engine.go
sparrc: I would prefer them here as there's no potential for them to be used anywhere else so no utility to the scope being package-wide

ecs-init/engine/engine.go
sharanyad: can you either make it a const or add a comment explaining the number?

sparrc: sure will do

ecs-init/engine/engine.go
jahkeup: :+1: to moving this to the outer scope, there's no need to _use_ them elsewhere in the package, but they also don't need to be in this scope (and it'll compile to roughly the same code). Please add these to the top with some comments for tuning later.

petderek: We've typically put all constants at the top as a style choice. It runs a bit against Effective Go's style recommendations, but its more or less what we've settled on.

FWIW, we also break their guidelines with variable names, where we've always prefered full words.

sparrc: `minTime` would need to become something like `minServiceStartRetryBackoffTime` to describe what a package-level constant is doing....is this preferable?

sparrc: I've gone ahead and made them package-level constants

ecs-init/engine/engine.go
sparrc: this is done now

ecs-init/engine/engine.go
```diff
@@ -33,6 +35,11 @@ const (
 	terminalSuccessAgentExitCode = 0
 	terminalFailureAgentExitCode = 5
 	upgradeAgentExitCode         = 42
+	minServiceStartRetryTime     = time.Millisecond * 500
```
jahkeup: nit: these constants should be prefixed the same as they're grouped to serve a common function.

```suggestion
	serviceMinStartRetryTime     = time.Millisecond * 500
```

ecs-init/engine/engine.go
```diff
@@ -180,15 +189,23 @@ func (e *Engine) StartSupervised() error {
...
 			err = e.upgradeAgent()
 			if err != nil {
 				log.Error("could not upgrade agent", err)
+			} else {
+				continue
 			}
+		case terminalFailureAgentExitCode:
+			return errors.New("agent exited with terminal exit code")
+		case terminalSuccessAgentExitCode:
+			return nil
```
jahkeup: There should be a `default` case to handle other exit codes not defined in this package - as the Agent is after a separate code base - to protect against looping over an unrecognized but failing-to-start/exiting-with-error Agent.

ecs-init/engine/engine.go
```diff
@@ -180,15 +189,23 @@ func (e *Engine) StartSupervised() error {
...
 		log.Infof("Agent exited with code %d", agentExitCode)
-		if agentExitCode == upgradeAgentExitCode {
+
+		switch agentExitCode {
+		case upgradeAgentExitCode:
 			err = e.upgradeAgent()
 			if err != nil {
 				log.Error("could not upgrade agent", err)
+			} else {
+				continue
```
jahkeup: I had to read through a few times to confirm that the intent with this `continue` is to intentionally start the Agent again and skip the retry given an upgraded image. Can we drop a comment here about restarting with the successfully upgraded agent? Is there reason to not try to start the agent as quickly as possible if it didn't upgrade successfully? Perhaps the upgrade handling logic could be simplified and handle this:

```go
if err != nil {
  log.Errorf("failed to upgrade agent: %s", err)
}
// start up the image loaded - upgraded or existing.
continue
```

---

It's very out of scope here but is however worth the thought exercise: the intent and handling would be clearer if the code was reasonable to tail-call itself in order to "start" again. ex:

```golang
upgradeErr := e.upgradeAgent()
if upgradeErr != nil {
  ... // log, handle "rollback" / stay-in-place start
}
return e.StartSupervised()
```

This is not going to happen here, but these handlers should be lower level concerns than what's going on with the supervision process.

ecs-init/engine/engine.go
```diff
@@ -33,6 +35,11 @@ const (
 	terminalSuccessAgentExitCode = 0
 	terminalFailureAgentExitCode = 5
 	upgradeAgentExitCode         = 42
+	minServiceStartRetryTime     = time.Millisecond * 500
+	maxServiceStartRetryTime     = time.Second * 15
+	serviceStartRetryJitter      = 0.10
+	serviceStartRetryMultiplier  = 2.0
+	serviceStartMaxRetries       = 999999999
```
jahkeup: This const *really* should be configurable, if nothing else then it should be `math.MaxInt64` to convey that we're not choosing an arbitrarily large number - its intended to be "forever". Perhaps add a comment regarding this decision?

ecs-init/engine/engine.go
```diff
@@ -180,15 +189,23 @@ func (e *Engine) StartSupervised() error {
...
 			}
+		case terminalFailureAgentExitCode:
+			return errors.New("agent exited with terminal exit code")
+		case terminalSuccessAgentExitCode:
+			return nil
 		}
-	}
-	if agentExitCode == terminalFailureAgentExitCode {
-		return errors.New("agent exited with terminal exit code")
+		d := retryBackoff.Duration()
```
jahkeup: If the Agent tickles any of the other cases, that don't cause `ecs-init` to exit, the `retryBackoff` could be excessive and delay the restart of the Agent unnecessarily. 

eg:

1. ECS init takes 3 retries to start up the Agent (`retries = 3`)
2. Agent signals upgrade (`retries = 3`)
3. Agent fails upgrade (`retries = 3`)
4. ECS init delays according to retryBackoff (`retries = 3`)
5. ECS init finally starts the Agent (`retries = 3`)

Can the `retryBackoff` be reset or expire after the Agent has been running for some period of time? Otherwise, customers with transient issues may be left with ECS init configured to *eventually* start the Agent but with a significant and inappropriate delay.

ecs-init/engine/engine.go
sparrc: Sure, I will give them the same prefix but I'm going to go with `serviceStart`

ecs-init/engine/engine.go
sparrc: all other exit codes should loop and hit the retry backoff, which is why there's no default case because we don't do anything on the default case.

ecs-init/engine/engine.go
sparrc: sure I will leave a comment

ecs-init/engine/engine.go
sparrc: yes I can use math.MaxInt64

ecs-init/engine/engine.go
sparrc: I'm not sure I follow. Each time ecs-init wants to start the agent it calls StartSupervised, this creates a new backoff object which will have the number of retries set to 0. Once the agent is running StartSupervised returns nil and the backoff counter is reset.

jahkeup: Each time that the loop runs, which in the case of an "upgrade exit" the loop isn't exited, the retry object is reused. So, in the listed scenario, you'd possibly have an Agent take the time at `retries = 3` to start, if that were to be made longer - maybe a longer lived instance or something - then you could have an Agent upgrade fail and take an unnecessarily long time to start up. 

I assume that the `serviceStartMaxRetryTime` is the upper bound and if so, its not a terribly large impact, but it is impact nonetheless.

ecs-init/engine/engine.go
jahkeup: Gotcha, I sorta glossed over that in my head :+1: It's concerning that we'd forever retry on errors we don't know anything about in this supervisor, but I suppose that issue is a larger one. 

The retry logic can be moved into the `default` case then which also mitigates the impact the retry would otherwise have on the failure and success cases of upgrading the Agent. What do you think about moving the retry to a `default:`?

ecs-init/engine/engine.go
```diff
@@ -33,6 +36,11 @@ const (
 	terminalSuccessAgentExitCode = 0
 	terminalFailureAgentExitCode = 5
 	upgradeAgentExitCode         = 42
+	serviceStartMinRetryTime     = time.Millisecond * 500
+	serviceStartMaxRetryTime     = time.Second * 15
+	serviceStartRetryJitter      = 0.10
+	serviceStartRetryMultiplier  = 2.0
+	serviceStartMaxRetries       = math.MaxInt64
```
jahkeup: Please add a comment on the intent of retrying forever. It clear this is a big number, but its a big number because its symbolic for "forever".

ecs-init/engine/engine.go
jahkeup: Yeah, I meant more that commentary for `continue` is surrounding its intent, which is to go straight to starting the upgraded Agent. The `if {} else {}` can go away either way.

ecs-init/engine/engine.go
sparrc: I didn't put it into a default because I don't think we want to get the agent in a situation where it's spamming many upgrade attempts per second.

ecs-init/engine/engine.go
sparrc: When it exits with "upgrade exit" exit code it will immediately try to upgrade the agent. If the upgrade is successful then the loop continues and doesn't increment the retry counter. The retry counter would only get incremented if the upgrade fails.

ecs-init/engine/engine.go
sparrc: sure will do

ecs-init/engine/engine.go
jahkeup: Yep, that's the scenario I outlined - where the upgrade fails. The point is that its possible to have a retry counter working even after a success case could have reset it.

ecs-init/engine/engine.go
jahkeup: Moving the retry handling (specifically the log line and the `sleep`) into the `default` case shouldn't be different from the logic that's there in the current proposed code. Is there another implication I'm not seeing?

sparrc: yep, the case is where the upgrade fails, in which case we _do_ want to trigger the retry backoff.

ecs-init/engine/engine.go
sparrc: The whole point of this PR is to put a delay in between retries, are you saying that the delay is fundamentally unnecessary in some cases? I guess I'd agree with you there but I'm not clear on how we would determine exactly when it's unnecessary. IMO the delay is not overly onerous as it starts at just 500ms and doubles each retry.

ecs-init/engine/engine.go
jahkeup: I see. 

I personally don't see why we would handle the failed upgrade case with a delay, as we aren't restarting due to an external problem that we're going to recover from differently if we delay.. but that said, if that's the design goal, happy to disagree and move forward.

ecs-init/engine/engine.go
jahkeup: Yeah, like I said, right now I think the delay isn't going to become too excessive. As I said above, I think the cases where the retry shouldn't compound are isolated from those where it should. If, for instance, the Agent were to be running for more than `serviceStartMaxRetryTime` then the retry could be reset without introducing a "too fast retry" situation, at least in the purview of `ecs-init`. 

I'm happy to disagree and move forward, these aren't blockers per se, but they should be ack'd and either noted for future improvement or consciously dropped.

sparrc: I'd prefer to leave it as it is. Introducing a running timer introduces a lot of complexity and it also can have unintended effects. If the agent takes some time to return an exit code, during which it was looping over some API calls that we are trying to limit, we might mistakenly assume that the agent has been running for a while and we are safe to run another retry.

That could circumvent what we are trying to achieve, because we don't know how many API calls the agent was running while it was trying to start, nor how long it tried to start for. Keeping the logic as-is enforces a downtime between retries that guarantees that we are not going to spam API calls to CP.

We would essentially just have to hope that the agent returns quickly in order for a running timer to work properly, which I don't think is a safe assumption.

PR #253, opened 2019-06-05
Add splunk to log driver defaults

petderek:
### Summary

This change adds splunk as a log driver available by default.

### Implementation details
"splunk" is added to the default environment variable.

This can be overridden if the ECS_AVAILABLE_LOGGING_DRIVERS variable is set in ecs.config.

```
# eg, if this is in the ecs.config file, splunk will not be enabled
ECS_AVAILABLE_LOGGING_DRIVERS=["awslogs"]
```

### Testing

WIP

New tests cover the changes: modified existing test

Add splunk to default log driver list

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -218,7 +218,7 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		"ECS_AGENT_CONFIG_FILE_PATH":            config.AgentJSONConfigFile(),
 		"ECS_UPDATE_DOWNLOAD_DIR":               config.CacheDirectory(),
 		"ECS_UPDATES_ENABLED":                   "true",
-		"ECS_AVAILABLE_LOGGING_DRIVERS":         `["json-file","syslog","awslogs","none"]`,
+		"ECS_AVAILABLE_LOGGING_DRIVERS":         `["json-file","syslog","awslogs","splunk","none"]`,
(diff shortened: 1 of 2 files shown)


sharanyad:
Should we also add fluentd?

tommyhahn:
I found an open PR related to adding fluentd by default:
https://github.com/aws/amazon-ecs-init/pull/151
Curious if there is an open/closed issue for the problem he mentioned?

sparrc:
closing because splunk logging driver requires docker api version 1.22+ and we support docker api 1.21+

see logging driver docker api requirements: https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/logging_drivers.go#L38 

and docker api version check in agent startup: https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent.go#L732

colinbjohnson:
@sparrc - should we put in an issue to support Docker API 1.22? Meaning: I feel like ECS will eventually support Docker API 1.22 at which point we should be able to use Splunk as a logdriver destination.

sparrc:
We support docker API 1.22, the problem is that we also support docker API version 1.21.

If we merged this, then the agent would fail to start on any instance using 1.21.

sparrc:
to clarify, we do support the splunk log driver, you just need to pass it in with `ECS_AVAILABLE_LOGGING_DRIVERS`. We shouldn't pass it as a default unless we deprecate support for docker api version 1.21.

PR #254, opened 2019-06-06
Update Agent to 1.29.0

yumex93:
### Testing

New tests cover the changes: 

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -31,7 +31,7 @@
 %endif
 
 Name:           ecs-init
-Version:        1.28.1
+Version:        1.29.0
 Release:        2%{?dist}
```
fenxiong: you should change this to 1

yumex93: Why this is 1? I think it should be 0.

Issue #259, opened 2019-09-20
Container labels are not available for configuration


morfien101:
### Summary

When running the ECS-Agent using AWS Linux for ECS you are not able to configure labels for the container. We use the labels for logging and grouping.

### Description

The container for ECS-Agent doesn't have the ability to set the docker container labels.

### Expected Behavior

There should be a mechanism to set the labels for the container. This can be injected at user-data boot time.

morfien101:
Are PRs and Issues not being looked at on this repo?

@sparrc Since you where the last person to commit something, can you answer the above question?

sparrc:
Thanks for the contribution and sorry for the delay looking at this, we had internal tickets that took priority. We are currently looking at this as a feature request and reviewing the PR, will let you know soon.

morfien101:
Hi @sparrc Any chance of an update on this PR. We would really like to see it get merged as its currently causing an issue for us.

morfien101:
:rolling_eyes: I pressed the wrong button and closed it. Its actually still open...

fenxiong:
This has been released in [v1.32.1](https://github.com/aws/amazon-ecs-init/releases/tag/v1.32.1-1)

PR #260, opened 2019-09-20
Adding in the ability to set container tags

morfien101:
Solves #259

A new Environment variable is available to ecs-init that will set the container labels when starting the container.

I choose to use a single layer JSON blob for the data. This is the same as how we set the logging drivers so it should be easy for admins to understand. If the decode fails it will log and ignore them. The is no error handling when starting so logging and continuing is the only option allowed by the application.

I'm not sure where to document this change other than on the ECS-Agent github page. I'm happy to submit a PR for that also if the change is accepted. Or someone can point me to the process of getting the page updated.


ecs-init/docker/docker.go
```diff
@@ -244,9 +246,16 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		env = append(env, envKey+"="+envValue)
 	}
 
+	lables, err := generateLabelMap(envVariables["ECS_CONTAINER_LABELS"])
```
fenxiong: this triggers error when ECS_CONTAINER_LABELS is not set:
```
2019-10-14T18:35:37Z [ERROR] Failed to decode the container labels, skipping labels. Error: unexpected end of JSON input
```
Can you skip generateLabelMap when ECS_CONTAINER_LABELS is empty so that we don't have an error log by default?

ecs-init/docker/docker.go
```diff
@@ -221,6 +222,7 @@ func (c *Client) getContainerConfig() *godocker.Config {
 		"ECS_AVAILABLE_LOGGING_DRIVERS":         `["json-file","syslog","awslogs","none"]`,
 		"ECS_ENABLE_TASK_IAM_ROLE":              "true",
 		"ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST": "true",
+		"ECS_CONTAINER_LABELS":                  "",
```
fenxiong: how about `ECS_AGENT_LABELS` since these labels are only for the agent container itself rather than all the containers managed by ECS

yumex93:
I will approve once you resolve @fenxiong's comments.

ecs-init/docker/docker_test.go
```diff
@@ -639,3 +639,29 @@ func TestStopAgent(t *testing.T) {
 		t.Error("Error should not be returned")
 	}
 }
+
+func TestContainerLabels(t *testing.T) {
```
sharanyad: Thanks for adding these unit tests. could you please add a case for empty / no labels?

morfien101:
I think I have done all the things requested in the flurry of activity. Its bed time in this part of the world. I'll check in again in the morning.

ecs-init/docker/docker.go
```diff
@@ -243,11 +245,28 @@ func (c *Client) getContainerConfig() *godocker.Config {
...
-	return &godocker.Config{
+	cfg := &godocker.Config{
 		Env:   env,
 		Image: config.AgentImageName,
 	}
+	setLabels(cfg, envVariables["ECS_AGENT_LABELS"])
+	return cfg
+}
+
+func setLabels(cfg *godocker.Config, lablesStringRaw string) {
```
fenxiong: a few typos in this function: lables -> labels

morfien101: :) I get caught with this all the time. UK based people want it different... Anyway its been fixed in the next commit.

morfien101:
Thank for the approval.

How long does it normally take to get shipped into a release?

We would like to schedule the work to make the changes in our configuration when this is generally available.

ellenthsu:
It generally takes a couple weeks for it to be generally available

shubham2892:
Can you squash your commits please.

Issue #227, opened 2019-01-04
`docker stop ecs-agent` does not prevent the agent from restarting on Amazon Linux 2


samuelkarp:
### Summary
`docker stop ecs-agent` does not prevent the agent from restarting on Amazon Linux 2

### Description
On Amazon Linux (and the old Upstart config), we had it set up so that normal exits (exit code of 0) would not cause Upstart to restart ecs-init (and consequently the agent).  This gave customers two paths to stopping: using `docker stop ecs-agent` (which meant they didn't need to know anything about Upstart) and `sudo stop ecs` (using Upstart).  It doesn't look like we preserved this behavior for Amazon Linux 2 with our systemd unit.

### Expected Behavior
```
$ docker stop ecs-agent
$ sudo systemctl status ecs.service
● ecs.service - ECS Agent
   Loaded: loaded (/usr/lib/systemd/system/ecs.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Fri 2019-01-04 19:36:38 UTC; 2s ago
```

### Observed Behavior
```
$ docker stop ecs-agent
$ sudo systemctl status ecs.service
● ecs.service - ECS Agent
   Loaded: loaded (/usr/lib/systemd/system/ecs.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Fri 2019-01-04 19:35:58 UTC; 4s ago
$ sudo systemctl status ecs.service
● ecs.service - ECS Agent
   Loaded: loaded (/usr/lib/systemd/system/ecs.service; enabled; vendor preset: disabled)
   Active: active (running) since Fri 2019-01-04 19:36:08 UTC; 3s ago
```

### Environment Details
ami-0c1f4871ebaae6d86

fierlion:
confirmed fixed with latest release https://github.com/aws/amazon-ecs-init/releases/tag/v1.35.0-1
output from an instance of the latest ecs-optimized AMI:
```
$ docker stop ecs-agent
ecs-agent
$ sudo systemctl status ecs.service
● ecs.service - Amazon Elastic Container Service - container agent
   Loaded: loaded (/usr/lib/systemd/system/ecs.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Fri 2019-12-20 16:38:14 UTC; 7s ago
```

PR #269, opened 2019-11-13
Don't fail the service when stopping a stopped agent

fenxiong:
### Summary

Fix https://github.com/aws/amazon-ecs-init/issues/227. The cause of the issue is that, on AL1, after the agent container stops, we don't try to stop it again (since it's in pre-stop), but on AL2, afterh the agent stops, we try to stop it again and fail (get a container not running error from Docker) since the pre-stop corresponds to the stop action of the systemd unit on AL2 (https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/ecs-init.go#L101-L104). The failure causes the service to be restarted.

Fixing this behavior on AL2 by not failing the stop step when the agent container is not running.

### Implementation details

Check stop container error from docker. if it's of type `ContainerNotRunning`, exits successfully.

### Testing

Unit test added (combined all the unit tests for `(c *Client) StopAgent()` together).

Manually built AL1 and AL2 rpm and verified that for both case: (1)  the agent remains stopped after stopping the job/service; (2) the agent remains stopped after the agent exits with 0.

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -423,5 +423,12 @@ func (c *Client) StopAgent() error {
 		return nil
 	}
 	stopContainerTimeoutSeconds := uint(10)
-	return c.docker.StopContainer(id, stopContainerTimeoutSeconds)
+	err = c.docker.StopContainer(id, stopContainerTimeoutSeconds)
+	if err != nil {
+		if _, ok := err.(*godocker.ContainerNotRunning); ok {
+			log.Info("Agent is already stopped")
+			return nil
+		}
+	}
+	return err
 }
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker.go
```diff
@@ -423,5 +423,12 @@ func (c *Client) StopAgent() error {
 		return nil
 	}
 	stopContainerTimeoutSeconds := uint(10)
-	return c.docker.StopContainer(id, stopContainerTimeoutSeconds)
+	err = c.docker.StopContainer(id, stopContainerTimeoutSeconds)
+	if err != nil {
+		if _, ok := err.(*godocker.ContainerNotRunning); ok {
+			log.Info("Agent is not running")
```
yunhee-l: in this case we should consider phrasing log entry differently to suggest that it's not an error case - if we just read "Agent is not running" in the log, that sounds bad. Something along the lines of "Agent is already stopped" or "Agent stopped successfully"?

fenxiong: yeah, changed to "Agent is already stopped"

fenxiong:
> lgtm.
> 
> nit: formatting seems to a little off. did you run gofmt on this?

Yeah i have, for the two files i touched

PR #273, opened 2019-12-17
capture a fixed tail of container logs when removing a container

fierlion:
Author:    fierlion <fierlion@amazon.com>
Date:      Fri Dec 6 00:21:02 2019 +0000
Committer: fierlion <fierlion@amazon.com>

### Summary

When the agent container fails, ecs-init removes the container as it starts up another.  We lose the failed container's logs.  This captures a set window of agent container logs and adds them to the ecs-init log for a failed container (exit code 2).

### Implementation details

In order to avoid possibly searching through a large log, this uses the --tail flag of the docker Log and a pre-configured window (here set to 200 lines).  This assures us consistent performance.  I added a panic() to the agent container binary to force the agent container failed exit.  This returned a '2' exit code.  
We can tweak the window and potentially the exit code(s) we capture logs for over time as necessary.

### Testing

I tested with a custom build agent with a panic in the StopContainer.  This allowed me to start an ecs task using ecs-cli, confirm that it was running using `docker ps` and then use the cli to stop the task which triggered the panic() in the agent and caused the agent container to fail.  I was able to capture logs from this failed container in `/var/log/ecs/ecs-init.log`.

New tests cover the changes: 
Yes

Capture a fixed tail of container logs when removing a container

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -14,6 +14,7 @@
 package docker
 
 import (
+	"bytes"
 	"encoding/json"
 	"io"
 	"path/filepath"
@@ -213,6 +214,31 @@ func (c *Client) StartAgent() (int, error) {
 		return 0, err
 	}
 	return c.docker.WaitContainer(container.ID)
+}
+
+// GetContainerLogTail will return the last logWindowSize lines of logs for
+// the Agent Container.
+func (c *Client) GetContainerLogTail(logWindowSize string) string {
+	containerToLog, _ := c.findAgentContainer()
+	if containerToLog == "" {
+		log.Info("No existing container to take logs from.")
+                return ""
+	}
+	// we want to capture some logs from our removed containers in case of failure
+	var containerLogBuf bytes.Buffer
+	err := c.docker.Logs(godocker.LogsOptions{
+		Container:    containerToLog,
+		OutputStream: &containerLogBuf,
+		Stdout:       true,
+		Stderr:       true,
+		Tail:         logWindowSize,
+		Timestamps:   true,
+	})
+	// we're ok if grabbing the container's logs fails
+	if err != nil {
+		log.Infof("Unable to tail logs for container ID: %s", containerToLog)
+	}
+	return containerLogBuf.String()
 }
 
 func (c *Client) getContainerConfig(envVarsFromFiles map[string]string) *godocker.Config {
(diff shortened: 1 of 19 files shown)


ecs-init/docker/docker.go
```diff
@@ -215,6 +216,30 @@ func (c *Client) StartAgent() (int, error) {
...
+// GetContainerLogTail will return the last logWindowSize lines of logs for
+// the Agent Container.
+func (c *Client) GetContainerLogTail(logWindowSize string) string {
+	containerToLog, _ := c.findAgentContainer()
+	if containerToLog == "" {
+		log.Info("No existing container to take logs from.")
+	}
+	// we want to capture some logs from our removed containers in case of failure
+	var containerLogBuf bytes.Buffer
+	err := c.docker.Logs(godocker.LogsOptions{
```
samuelkarp: This API requires a Docker log driver that supports reading logs.  Not all of them do.  Even though we don't explicitly specify the log driver when starting the agent container, customers can set the log driver at the daemon level.  You'll need to expect failures to happen in cases where customers have configured a different default log driver.

fierlion: Agree.  I believe this is a good starting point for the set of cases without a different default log driver.  Presently we have no visibility into these failed (and reaped) agent containers.

petderek: > This API requires a Docker log driver that supports reading logs. Not all of them do. 

This will work in the default case, and fail gracefully other cases. It's not perfect, but it still means we capture failures in a few more failure scenarios.

IMO the long term fix would be to have agent no longer manage its own log file. That change is much more invasive to existing setups, however.

ecs-init/docker/docker.go
```diff
@@ -215,6 +216,30 @@ func (c *Client) StartAgent() (int, error) {
 	return c.docker.WaitContainer(container.ID)
 }
 
+// GetContainerLogTail will return the last logWindowSize lines of logs for
+// the Agent Container.
+func (c *Client) GetContainerLogTail(logWindowSize string) string {
+	containerToLog, _ := c.findAgentContainer()
+	if containerToLog == "" {
+		log.Info("No existing container to take logs from.")
```
sparrc: should we return here?

fierlion: yep.  Added.

PR #274, opened 2019-12-19
initial PR for Amazon ECS Volume Plugin

sharanyad:
### Summary
This PR contains the interface and implementation skeleton for ECS volume plugin. 
It is currently a WIP. vendor updates to follow. 

### Testing

```
PASS
coverage: 90.1% of statements
ok  	github.com/aws/amazon-ecs-init/ecs-init/volumes	0.005s	coverage: 90.1% of statements
```

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -1,0 +1,92 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package volumes
+
+import (
+	"fmt"
+	"strconv"
+	"sync"
+)
+
+// ECSVolumeDriver holds mount helper and methods for different Volume Mounts
+type ECSVolumeDriver struct {
+	volumeMounts map[string]*MountHelper
+	lock         sync.RWMutex
+}
+
+// NewECSVolumeDriver initializes fields for volume mounts
+func NewECSVolumeDriver() *ECSVolumeDriver {
+	return &ECSVolumeDriver{
+		volumeMounts: make(map[string]*MountHelper),
+	}
+}
+
+// Create implements ECSVolumeDriver's Create volume method
+func (e *ECSVolumeDriver) Create(r *CreateRequest) error {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+
+	if _, ok := e.volumeMounts[r.Name]; ok {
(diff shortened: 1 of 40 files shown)


ecs-init/volumes/ecs_volume_driver.go
```diff
@@ -0,0 +1,238 @@
...
+
+// AmazonECSVolumeDriver holds list of volume drivers and volumes information
+type AmazonECSVolumeDriver struct {
+	volumeDrivers map[string]ECSVolumeDriver
+	volumes       map[string]*Volume
+	lock          sync.RWMutex
+}
+
+// ECSVolumeDriver contains the methods for volume drivers to implement
+type ECSVolumeDriver interface {
```
petderek: Currently we have a VolumeDriver that calls more VolumeDrivers. IMO, I'd rather call this abstraction layer something else. MountableVolume, MountHelper, Mount, etc

ecs-init/volumes/efs_driver.go
```diff
@@ -0,0 +1,91 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package volumes
```
petderek: consider getting rid of the efs* descriptions and leaving this as a generic mount helper for now. As written, this could be reused for any mount helper installed on the host.

MB we have this be the default case, and then add EFS specific logic to a different struct via embedding. Just a thought.

ecs-init/volumes/ecs_volume_driver.go
sharanyad: we already have a mount helper as part of the struct (`EFSVolumeDriver` ) that implements this interface, hence didn't want to call this MountHelper. To me, it makes sense that the main volume driver is composed of multiple drivers for different volume types. How about calling the main one `AmazonECSVolumePlugin` instead of `AmazonECSVolumeDriver` ?

ecs-init/volumes/efs_driver.go
sharanyad: called it generically as "ECS Volume Driver" which has only EFS mount helper as of now. we can make it an interface in the future to have different mount helpers :)

ecs-init/volumes/efs_mount_helper.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := m.Validate(); err != nil {
+		return err
+	}
+	return RunMount(args)
+}
+
+var RunMount = RunMountCommand
+
+func RunMountCommand(args []string) error {
```
petderek: these probably don't need to be exported

sharanyad: ah yeah, used in the same package! will change

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -0,0 +1,249 @@
...
+
+// VolumeDriver contains the methods for volume drivers to implement
+type VolumeDriver interface {
+	Create(*CreateRequest) error
+	Remove(*RemoveRequest) error
+}
+
+// Volume holds full details about a volume
+type Volume struct {
+	Type    string
```
yumex93: Is type going to just be efs for now?

PR #277, opened 2019-12-28
store state information of the plugin to disk

sharanyad:
### Summary
Amazon ECS Volume Plugin writes state information of the volumes
that it manages to `/var/lib/ecs/data/ecs_volume_plugin.json`. On
startup, it loads the volumes information from the disk and periodically
saves state on each volume's successful create and delete operations.
Tested manually.
state file after creating volume `efsVolume456` : 
```
$ sudo cat /var/lib/ecs/data/ecs_volume_plugin.json
{
	"volumes": {
		"efsVolume456": {
			"type": "efs",
			"path": "/var/lib/ecs/volumes/efsVolume456",
			"options": {
				"device": "fs-1cf5f8b7",
				"o": "tls",
				"type": "efs"
			}
		}
	}
}
```
stopped the plugin and restarted 
```
$ docker volume ls
amazon-ecs-volume-plugin   efsVolume456
```
volume removal:
```
$ docker volume rm efsVolume456

$ sudo cat /var/lib/ecs/data/ecs_volume_plugin.json
{}

```

### Testing

### Licensing

### DIFF
--- a/ecs-init/volumes/amazon-ecs-volume-plugin/plugin.go
+++ b/ecs-init/volumes/amazon-ecs-volume-plugin/plugin.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"os"
 	"os/user"
 	"strconv"
 
@@ -23,6 +24,9 @@ import (
 
 func main() {
 	plugin := volumes.NewAmazonECSVolumePlugin()
+	if err := plugin.LoadState(); err != nil {
+		os.Exit(1)
+	}
 	handler := volume.NewHandler(plugin)
 	rootUser, _ := user.Lookup("root")
 	gid, _ := strconv.Atoi(rootUser.Gid)
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -33,21 +33,25 @@ const (
 type AmazonECSVolumePlugin struct {
 	volumeDrivers map[string]VolumeDriver
 	volumes       map[string]*Volume
+	state         *StateManager
 	lock          sync.RWMutex
 }
 
 // NewAmazonECSVolumePlugin initiates the volume drivers
 func NewAmazonECSVolumePlugin() *AmazonECSVolumePlugin {
-	return &AmazonECSVolumePlugin{
+	plugin := &AmazonECSVolumePlugin{
 		volumeDrivers: map[string]VolumeDriver{
 			"efs": NewECSVolumeDriver(),
 		},
 		volumes: make(map[string]*Volume),
+		state:   NewStateManager(),
 	}
+	return plugin
 }
 
 // VolumeDriver contains the methods for volume drivers to implement
 type VolumeDriver interface {
+	Setup(string, *Volume)
 	Create(*CreateRequest) error
 	Remove(*RemoveRequest) error
 }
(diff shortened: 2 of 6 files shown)


ecs-init/volumes/state_manager.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+)
+
+const (
+	// PluginStatePath is the directory path to the plugin state information
+	PluginStatePath = "/var/lib/ecs/data/"
```
petderek: you should add a TODO to have this observe ECS_DATA_DIR

ecs-init/volumes/state_manager.go
```diff
@@ -0,0 +1,101 @@
...
+	// PluginStateFile contains the state information of the plugin
+	PluginStateFile = "ecs_volume_plugin.json"
+	// PluginStateFileAbsPath is the absolute path of the plugin state file
+	PluginStateFileAbsPath = "/var/lib/ecs/data/ecs_volume_plugin.json"
+)
+
+var lock sync.Mutex
+
+// SaveState saves a representation of 'a' to the file at path
+func SaveState(a interface{}) error {
```
petderek: I think we should have this method take a strongly typed object instead of relying on the Volume object. The data persistence layer should have a different model. The two benefits we get from this:

1) you don't need to export all of the fields of Volume in the main method
2) we don't need to worry about how changing Volume in future impacts serialization -- they're decoupled

The downside here is that we need to make changes in two places if we want more data to be persisted -- but, this may also be desirable as it is more explicit. IMO, its worth it given our experiences with managing state in agent.

ecs-init/volumes/state_manager.go
```diff
@@ -0,0 +1,101 @@
...
+const (
+	// PluginStatePath is the directory path to the plugin state information
+	PluginStatePath = "/var/lib/ecs/data/"
+	// PluginStateFile contains the state information of the plugin
+	PluginStateFile = "ecs_volume_plugin.json"
+	// PluginStateFileAbsPath is the absolute path of the plugin state file
+	PluginStateFileAbsPath = "/var/lib/ecs/data/ecs_volume_plugin.json"
+)
+
+var lock sync.Mutex
```
petderek: having a package level lock is a little scary. We should probably bring this into  a state saver object.

sharanyad: that was my initial thought too. 
two things considered:
1. the struct will only have lock as the field
2. since initially Volume plugin interface was going to be persisted as is, this object would be a global var.

But I can change this to having a different model for persistence and have a state manager object as well as part of the volume plugin.

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -125,6 +165,12 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 
 	// record the volume information
 	a.volumes[r.Name] = vol
+
+	// save the state of new volume
+	err = a.state.recordVolume(r.Name, vol)
```
petderek: this simplifies to return a.state.recordVolume(...)

ecs-init/volumes/amazon-ecs-volume-plugin/plugin.go
```diff
@@ -23,6 +24,9 @@ import (
 
 func main() {
 	plugin := volumes.NewAmazonECSVolumePlugin()
+	if err := plugin.LoadState(); err != nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
sharanyad: sure, will add it to the PR that has plugin logging

ecs-init/volumes/ecs_volume_plugin.go
sharanyad: yeah, in the other PR, I log something if err != nil, hence this path 
(https://github.com/aws/amazon-ecs-init/pull/283/files#diff-c11b0b2ff52442451c3df2b3f2380e61R185)

PR #284, opened 2020-01-08
Release 1.36.0-1

sparrc:
### Summary

ecs init release 1.36.0

- Cache Agent version 1.36.0
- Capture a fixed tail of container logs when removing a container

### Licensing


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -255,6 +255,11 @@ fi
 %endif
 
 %changelog
+
```
shubham2892: Version of line 35 also needs to be updated.

sparrc: done

Issue #182, opened 2018-07-05
SUSE systemd unit missing appropriate stop, post-stop directives


jahkeup:
### Summary

The SUSE systemd unit is missing the directives now present and accounted for in the AL2 systemd unit.

fierlion:
https://github.com/aws/amazon-ecs-init/blob/master/packaging/suse/amazon-ecs.service

fierlion:
https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs.service#L29-L30

jahkeup:
@fierlion not sure what's meant with these links, I think there's still differences present here though I'm somewhat removed from on going work in `ecs-init` at this point. From a quick glance, I do see that we're not calling the same subcommands that we're calling and using in AL2 (which was what this issue was opened to address). We should test/get-confirmations that the service works correctly and behaves similarly on OpenSUSE as it does in AL2.

fierlion:
Sorry.  See updates to my comment above.

jahkeup:
Okay, I got it :+1: 

So the real outstanding task here is to actually go and test it before we commit that change (which I think was the reason why I deferred the effort to an issue long, long ago..).

samuelkarp:
https://build.opensuse.org/package/show/openSUSE:Factory/amazon-ecs-init appears to be the place where the OpenSUSE build is produced.  @rjschwei is the maintainer who had been helping us package ecs-init for that distro.

fierlion:
> So the real outstanding task here is to actually go and test it before we commit that change (which I think was the reason why I deferred the effort to an issue long, long ago..), is that right?

that's right.  I'll update here with my manual test results on SUSE.

fierlion:
https://github.com/aws/amazon-ecs-init/pull/285

rjschwei:
https://build.opensuse.org/package/show/openSUSE:Factory/amazon-ecs-init is the derived package for release with openSUSE Tumbleweed which will eventually make it's way into openSUSE Leap and SLE. Development of the package happens here:

https://build.opensuse.org/package/show/Cloud:Tools/amazon-ecs-init

fierlion:
Merged.  I'll open a follow-up issue to work through the `make rpm` target and the suse spec file.  `make rpm` is pinned to amazon-linux and the suse spec file hasn't been updated in a while and requires attention.

PR #285, opened 2020-01-10
Update suse systemd Unit file

fierlion:
### Summary

This adds the appropriate stop, post-stop directives as requested here:
https://github.com/aws/amazon-ecs-init/issues/182

### Implementation details

The `pre-stop` and `stop` actions point to the same `engine.PreStop` function
https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/ecs-init.go#L97-L104
I updated the ExecStop to point to `stop` so it mirrors what we have in amazon-linux-ami
https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs.service#L29-L30
I also added the ExecStopPost directive.  
See testing for more detail.
### Testing

1. Started a new EC2 instance from the `SUSE Linux Enterprise Server 15 SP1 (HVM), SSD Volume Type - ami-0e6612c7b620ecf3a` AMI in us-west-2.  
2. Set up Docker, provisioned host with necessary RPMs and then pulled the latest ECS-agent image from DockerHub.  
3. Cloned amazon-ecs-init and made the updates included here in the SUSE Unit file
4. Built ecs-init on the host directly using the above ecs-agent container with `make rpm`.  
4.  Installed the rpm using `sudo rpm -i <rpm file>`
5. Manually copied the new unit file into /etc/systemd/system and the amazon-ecs-init binary into `/usr/sbin/` and ran `sudo systemctl enable /etc/systemd/system/amazon-ecs.service`
6. Started up ecs-agent container using `sudo systemctl start amazon-ecs`.  Waited for the agent container to become healthy.
7. Stopped the ecs-agent container with `sudo systemctl stop amazon-ecs`
8. Verified `/var/log/ecs/ecs-init.log` contained the expected INFO output from the new post-stop step:
```
2020-01-10T18:23:33Z [INFO] post-stop
2020-01-10T18:23:33Z [INFO] Cleaning up the credentials endpoint setup for Amazon Elastic Container Service Agent
```

### Licensing

### DIFF
--- a/packaging/suse/amazon-ecs.service
+++ b/packaging/suse/amazon-ecs.service
@@ -10,7 +10,8 @@ Requires=network.target
 Type=simple
 ExecStartPre=/usr/sbin/amazon-ecs-init pre-start
 ExecStart=/usr/sbin/amazon-ecs-init start
-ExecStop=/usr/sbin/amazon-ecs-init pre-stop
+ExecStop=/usr/sbin/amazon-ecs-init stop
+ExecStopPost=/usr/sbin/amazon-ecs-init post-stop
 ExecReload=/usr/sbin/amazon-ecs-init reload-cache
 
 [Install]


samuelkarp:
> Manually copied the new unit file into /etc/systemd/system and the amazon-ecs-init binary into /usr/sbin/ [...]

Why was this step necessary?  These files should be installed as part of installing the RPM.  See [here](https://github.com/aws/amazon-ecs-init/blob/7216f783b7dbf3e0641c554a6b2091fb9c5aec59/packaging/suse/amazon-ecs-init.spec#L53) and [here](https://github.com/aws/amazon-ecs-init/blob/master/packaging/suse/amazon-ecs-init.spec#L56).

fierlion:
> Why was this step necessary

`make RPM` seems to use the amazon linux rpm spec.  It didn't do what I wanted, but it did execute to completion.  That's the only reason I mentioned it in the steps.
I knew that if I put my systemd Unit file and its associated amazon-ecs-init binary in the right places, this would do all I needed to test if these specific unit file changes worked.  
If there's further testing warranted, let me know.

 > @rjschwei is the maintainer who had been helping us package ecs-init 

Note: I've reached out for a code review from @rjschwei on this.  I won't merge without all the necessary feedback.

rjschwei:
Thanks for the heads up LGTM

PR #287, opened 2020-01-14
accept target field for volume creation

sharanyad:
### Summary
Earlier, target field was ignored for volume creation and mount. Now if target path is specified, the volume is mounted at that location. 

### Testing

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -129,11 +129,13 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 	}
 
 	// get driver type from options to get the corresponding volume driver
-	var driverType string
+	var driverType, target string
 	for k, v := range r.Options {
 		switch k {
 		case "type":
 			driverType = v
+		case "target":
+			target = v
 		}
 	}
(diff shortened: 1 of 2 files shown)


ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -129,11 +129,13 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 	}
 
 	// get driver type from options to get the corresponding volume driver
-	var driverType string
+	var driverType, target string
```
yumex93: Does this mean customer can specify a target path to specify where they want to mount the volume? Or agent is the one who decides the target?

sharanyad: for EFS, agent doesnt pass in target and plugin creates the mount path. 
This is just added for mostly testing purposes

yumex93: What kind of testing?

sharanyad: if we don't want to mess with the original ECS volumes dir. example: the mount path created by plugin is /var/lib/ecs/volumes/<volume-name> . I can override it using the target field and have /foo as mount

PR #292, opened 2020-01-16
not failing create/remove calls if save state to disk fails

sharanyad:
### Summary
In the current code, after successfully creating/mounting/removing volumes, if saving the volume info to disk fails, we return an error to Docker. This is not correct, since the volumes are actually successfully created/removed. 

The state will be saved to disk again during the next create/remove call. 

### Testing

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -185,8 +185,7 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 	// save the state of new volume
 	err = a.state.recordVolume(r.Name, vol)
 	if err != nil {
-		seelog.Errorf("Error saving state of new volume %s", r.Name)
-		return err
+		seelog.Errorf("Error saving state of new volume %s: %v", r.Name, err)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -275,18 +274,17 @@ func (a *AmazonECSVolumePlugin) Remove(r *volume.RemoveRequest) error {
...
 	// cleanup the volume's host mount path
 	err = a.CleanupMountPath(vol.Path)
 	if err != nil {
 		seelog.Warnf("Cleaning mount path failed for volume %s: %v", r.Name, err)
 	}
+	seelog.Infof("Saving state after removing volume %s", r.Name)
+	// remove the state of deleted volume
+	err = a.state.removeVolume(r.Name)
+	if err != nil {
+		seelog.Errorf("Error saving state after removing volume %s", r.Name)
```
fenxiong: i think it's better to log the error?

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -275,18 +274,17 @@ func (a *AmazonECSVolumePlugin) Remove(r *volume.RemoveRequest) error {
...
 	delete(a.volumes, r.Name)
-	seelog.Infof("Saving state after removing volume %s", r.Name)
-	// remove the state of deleted volume
-	err = a.state.removeVolume(r.Name)
-	if err != nil {
-		seelog.Errorf("Error saving state")
-		return err
-	}
 	// cleanup the volume's host mount path
 	err = a.CleanupMountPath(vol.Path)
```
fenxiong: should this error be returned? seems like you are ignoring it as well

ecs-init/volumes/ecs_volume_plugin.go
sharanyad: thats right, will log it.

ecs-init/volumes/ecs_volume_plugin.go
sharanyad: no, this is same as state save. as long as unmount is successful, we should be good

petderek: We should probably log an error here. If something goes wrong and we can't cleanup: we need to figure out if we need to leave the volume in place or not. 

I'm leaning toward removing the volume, or maybe adding a retry with a force flag depending on how testing works out.  Agent will ensure that dependent processes stopped before destroying the volume.

ecs-init/volumes/ecs_volume_plugin.go
petderek: what do you mean in this case?

ecs-init/volumes/ecs_volume_plugin.go
sharanyad: will change warnf to errorf. also this just removes the volume mount path. even if its not removed, a new volume with the same volume name can still be mounted to the same path. 

with respect to adding retries for remove/unmount volume failure, i prefer not to do that and follow what docker does , i.e failing it. there could be various reasons such as some other container is consuming the volume/mount helper not running etc, which plugin cannot handle.

ecs-init/volumes/ecs_volume_plugin.go
sharanyad: earlier the error was not logged. i have changed it.

PR #293, opened 2020-01-17
return mount path and createdAt for get volume API

sharanyad:
### Summary
To get information such as `createdAt` time, mount point for the volume, return it from the plugin. 

Testing:
```
$ docker volume inspect efsvolume0
[
    {
        "CreatedAt": "2020-01-17T21:33:38Z",
        "Driver": "amazon-ecs-volume-plugin",
        "Labels": {},
        "Mountpoint": "/var/lib/ecs/volumes/efsvolume0",
        "Name": "efsvolume0",
        "Options": {
            "device": "fs-1cf5f8b7",
            "o": "tls,ro",
            "type": "efs"
        },
        "Scope": "local"
    }
]
```
State file:
```
$ sudo cat /var/lib/ecs/data/ecs_volume_plugin.json
{
	"volumes": {
		"efsvolume0": {
			"type": "efs",
			"path": "/var/lib/ecs/volumes/efsvolume0",
			"options": {
				"device": "fs-1cf5f8b7",
				"o": "tls,ro",
				"type": "efs"
			},
			"createdAt": "2020-01-17T21:33:38.172301878Z"
		}
	}
}
```

### Testing

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/cihub/seelog"
 	"github.com/docker/go-plugins-helpers/volume"
@@ -59,9 +60,10 @@ type VolumeDriver interface {
 
 // Volume holds full details about a volume
 type Volume struct {
-	Type    string
-	Path    string
-	Options map[string]string
+	Type      string
+	Path      string
+	Options   map[string]string
+	CreatedAt string
 }
 
 // CreateRequest holds fields necessary for creating a volume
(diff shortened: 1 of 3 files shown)


ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -310,14 +313,17 @@ func (a *AmazonECSVolumePlugin) List() (*volume.ListResponse, error) {
...
 	}
-	vol := &volume.Volume{
-		Name: r.Name,
+	resp := &volume.Volume{
+		Name:       r.Name,
+		Mountpoint: vol.Path,
+		CreatedAt:  vol.CreatedAt,
 	}
-	return &volume.GetResponse{Volume: vol}, nil
+	seelog.Infof("Returning volume information: %v", resp)
```
sparrc: how frequently are we expecting to call this function? Do we need to log the full response every time?

yunhee-l: on similar note, maybe consider extracting minimum necessary info from resp to log?

sharanyad: just tested. this is printed twice for mounting volume to a container, docker stop container and docker volume inspect. This could become noisy if we are printing the full response. But this could still be valuable when debugging failures. I'll change it to it prints just the volume name.

PR #295, opened 2020-01-23
Add script to generate/update changelogs

fierlion:
### Summary
Currently we maintain multiple changelogs in ecs-init manually.  The release process makes updating the changelogs inconsistent. 
I'm seeking to make the process automatic. 
This creates a core changelog in the CHANGE_TEMPLATE file with all required details for each of the different changelog formats: amazon-linux/debian/markdown.
For each release, the engineer will update the CHANGE_TEMPLATE then will execute the changelog.go file (`go build changelog.go && ./changelog`)
Running the included script will update all the changelogs in place with the relevant contents of the CHANGE_TEMPLATE.

NOTE: All included new changelogs have been generated by the script.

### Implementation details
Each or the entries in the CHANGE_TEMPLATE is parsed into a `Change` object.  An array of these change objects is passed to each of the formatters to generate a new changelog string.  Finally, the changelog files are overwritten with these updated changelog strings.

### Testing
I used the script and `git diff` to build toward a minimal diff between the current changelogs and the generated changelogs.  Tested manually to generate all changelogs.

I also successfully ran `make rpm` from an AL2 ecs-optimized instance using the generated changelog for amazon-linux.

Add script to generate changelogs from a central file

### Licensing

### DIFF
--- a/scripts/changelog/changelog.go
+++ b/scripts/changelog/changelog.go
@@ -1,0 +1,208 @@
+package main
+
+import (
+    "bufio"
+    "fmt"
+    "os"
+    "strings"
+    "time"
+)
+
+type Change struct {
+    Version string
+    Name string
+    Datetime time.Time
+    Changes []string
+}
+
+const (
+    CHANGE_TEMPLATE = "./CHANGELOG_MASTER"
+    TOP_LEVEL = "../../CHANGELOG.md"
+    AMAZON_LINUX = "../../packaging/amazon-linux-ami/ecs-init.spec"
+    SUSE = "../../packaging/suse/amazon-ecs-init.changes"
+    UBUNTU = "../../packaging/ubuntu-trusty/debian/changelog"
+
+    AMAZON_LINUX_TIME_FMT = "Mon Jan 02 2006"
+    DEBIAN_TIME_FMT = "Mon, 02 Jan 2006 15:04:05 -0700"
+    SUSE_TIME_FMT = "Mon Jan 02, 15:04:05 MST 2006"
+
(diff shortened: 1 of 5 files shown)


scripts/changelog/changelog.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+    CHANGE_TEMPLATE = "./CHANGE_TEMPLATE"
+    TOP_LEVEL = "../../CHANGELOG.md"
+    AMAZON_LINUX = "../../packaging/amazon-linux-ami/ecs-init.spec"
+    SUSE = "../../packaging/suse/amazon-ecs-init.spec"
+    UBUNTU = "../../packaging/ubuntu-trusty/debian/changelog"
+
+    AMAZON_LINUX_TIME_FMT = "Mon Jan 02 2006"
+    DEBIAN_TIME_FMT = "Mon, 02 Jan 2006 15:04:05 -0700"
+    SUSE_TIME_FMT = "Mon Jan 02, 15:04:05 MST 2006"
```
yumex93: Does this get used anywhere? We don't need to add changelog for SUSE?

fierlion: We will add changelog for SUSE when we update the SUSE rpm process.  I can remove this SUSE-specific line if necessary.  Originally I was planning on doing all three changelogs at once, but the SUSE packaging is very out of date and will need some work.

fierlion:
NOTE: I based SUSE changes on https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)

PR #297, opened 2020-01-27
volume plugin capability for EFS and unit tests

sharanyad:
### Summary
This env var will be used by agent to send capability for EFS support by ECS Volume Plugin. 
Using the capability "efsAuth" and not "efs", since preview already supports "efs" capability , so to differentiate from that, we are using the "efsAuth" as the value. 

Added unit tests for efs_mount_helper.go , now that netns and creds uri options are finalized. 

### Testing

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -222,7 +222,7 @@ func (c *Client) GetContainerLogTail(logWindowSize string) string {
 	containerToLog, _ := c.findAgentContainer()
 	if containerToLog == "" {
 		log.Info("No existing container to take logs from.")
-                return ""
+		return ""
 	}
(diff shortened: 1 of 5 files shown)


ecs-init/docker/docker.go
```diff
@@ -222,7 +222,7 @@ func (c *Client) GetContainerLogTail(logWindowSize string) string {
 	containerToLog, _ := c.findAgentContainer()
 	if containerToLog == "" {
 		log.Info("No existing container to take logs from.")
-                return ""
+		return ""
```
petderek: do we not auto gofmt this stuff?

sharanyad: looks like we do not in init.

PR #305, opened 2020-02-19
Fix ECS_UPDATE_DOWNLOAD_DIR functionality

sparrc:
### Summary

Customers can use ECS_UPDATE_DOWNLOAD_DIR to tell the agent to download
updates to a custom directory, rather than the agent cache directory at
/var/cache/ecs. In order for the upgrade to take place, ecs-init also
needs access to this directory and needs to know about it by reading the
environment variable.

### Testing

An upgrade using the default value (/var/cache/ecs) and setting a custom value set using ECS_UPDATE_DOWNLOAD_DIR  was tested manually. New unit tests have also been added.

Fix ECS_UPDATE_DOWNLOAD_DIR functionality

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -135,6 +135,14 @@ func CacheDirectory() string {
 	return directoryPrefix + "/var/cache/ecs"
 }
 
+// UpdateDownloadDir returns to location of the agent update download dir
+func UpdateDownloadDir() string {
+	if downloadDir := os.Getenv("ECS_UPDATE_DOWNLOAD_DIR"); downloadDir != "" {
+		return downloadDir
+	}
+	return CacheDirectory()
+}
+
 // CacheState returns the location on disk where cache state is stored
 func CacheState() string {
 	return CacheDirectory() + "/state"
@@ -165,7 +173,7 @@ func AgentRemoteTarballMD5Key() (string, error) {
 
 // DesiredImageLocatorFile returns the location on disk of a well-known file describing an Agent image to load
 func DesiredImageLocatorFile() string {
-	return CacheDirectory() + "/desired-image"
+	return UpdateDownloadDir() + "/desired-image"
 }
 
 // DockerUnixSocket returns the docker socket endpoint and whether it's read from DockerHostEnvVar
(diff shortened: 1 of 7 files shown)


ecs-init/docker/docker.go
```diff
@@ -371,6 +371,10 @@ func (c *Client) getHostConfig(envVarsFromFiles map[string]string) *godocker.Hos
 		// bind mount instance config dir
 		config.InstanceConfigDirectory() + ":" + config.InstanceConfigDirectory(),
 	}
+	// only add update download dir if it is not the cache directory (the default setting)
+	if config.UpdateDownloadDir() != config.CacheDirectory() {
+		binds = append(binds, config.UpdateDownloadDir()+":"+config.UpdateDownloadDir())
```
fenxiong: do we need to change the container path to be the same? if so, seems like customer has to specify the download dir to be an existing dir on agent's file system?

ecs-init/config/common.go
```diff
@@ -135,6 +135,14 @@ func CacheDirectory() string {
 	return directoryPrefix + "/var/cache/ecs"
 }
 
+// UpdateDownloadDir returns to location of the agent update download dir
+func UpdateDownloadDir() string {
+	if downloadDir := os.Getenv("ECS_UPDATE_DOWNLOAD_DIR"); downloadDir != "" {
```
fenxiong: for AL2, the envs seen by ecs-init process contains the ones in /etc/ecs/ecs.config because of [this line](https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs.service#L26). But for AL1, IIUC i don't think they can be retrieved from os.Getenv and I think you need to get the value from [here](https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/docker/docker.go#L203)

sparrc: good catch, fixing

samuelkarp:
The purpose of `ECS_UPDATE_DOWNLOAD_DIR` is to configure the location of the downloaded update *inside* the agent container, not to configure the location on the host.

There are multiple assumptions in ecs-init about the location of updates, including the mounts that are added to the agent container (which is handled in this PR) and in the RPM spec (not handled in this PR and not really changeable).  If this change is made, installing a new version of ecs-init through yum will not properly upgrade the agent; the RPM will install the new agent to the `/var/cache/ecs` and not to whatever location is defined in `ECS_UPDATE_DOWNLOAD_DIR`.

I would recommend against making this change and instead discouraging customers from setting `ECS_UPDATE_DOWNLOAD_DIR`; it's an implementation detail of the coordination between ecs-init and the ECS agent rather than an intentionally configurable setting.

sparrc:
closing (see https://github.com/aws/amazon-ecs-agent/pull/2373)

PR #306, opened 2020-02-22
docker: add '/etc/alternatives' to mounts

petderek:
### Summary

In some circumstances, /sbin/iptables may be a symlink to a binary in
/etc/alternatives. In order to handle this case, its imperative that
agent is also able to access the /etc/alternatives folder.

### Implementation details

We now mount in "/etc/alternatives" so that binaries within that directory are callable by agent.

### Testing

**Testing is still WIP**

1. i build an rpm using `make rpm`
2. I installed it on an AL2 AMI
3. I've validated that the following appears upon docker inspect:

```
            "Binds": [
                "/var/run:/var/run",
                "/var/log/ecs:/log",
                "/var/lib/ecs/data:/data",
                "/etc/ecs:/etc/ecs",
                "/var/cache/ecs:/var/cache/ecs",
                "/sys/fs/cgroup:/sys/fs/cgroup",
                "/var/lib/ecs:/var/lib/ecs",
                "/etc/pki:/etc/pki:ro",
                "/run/docker/plugins:/run/docker/plugins:ro",
                "/etc/docker/plugins:/etc/docker/plugins:ro",
                "/usr/lib/docker/plugins:/usr/lib/docker/plugins:ro",
                "/proc:/host/proc:ro",
                "/usr/lib:/usr/lib:ro",
                "/lib:/lib:ro",
                "/usr/lib64:/usr/lib64:ro",
                "/lib64:/lib64:ro",
                "/sbin:/host/sbin:ro",
                "/etc/alternatives:/etc/alternatives:ro",
                "/usr/sbin:/usr/sbin:ro"
            ],


```

Bug - Fix agent's use of iptables when installed as an alternative

### Licensing

### DIFF
--- a/ecs-init/docker/docker_test.go
+++ b/ecs-init/docker/docker_test.go
@@ -29,8 +29,8 @@ import (
 // Note: Change this value every time when a new bind mount is added to
 // agent for the tests to pass
 const (
-	expectedAgentBindsUnspecifiedPlatform = 17
-	expectedAgentBindsSuseUbuntuPlatform  = 15
+	expectedAgentBindsUnspecifiedPlatform = 19
+	expectedAgentBindsSuseUbuntuPlatform  = 17
 )
 
 var expectedAgentBinds = expectedAgentBindsUnspecifiedPlatform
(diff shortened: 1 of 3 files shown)


ecs-init/docker/docker_test.go
```diff
@@ -29,8 +29,8 @@ import (
 // Note: Change this value every time when a new bind mount is added to
 // agent for the tests to pass
 const (
-	expectedAgentBindsUnspecifiedPlatform = 17
-	expectedAgentBindsSuseUbuntuPlatform  = 15
+	expectedAgentBindsUnspecifiedPlatform = 18
```
yumex93: The test also needs to be updated to check the new added bind. The expectKey part.

petderek: will do

petderek:
Adding both `/etc/alternatives/` and `/usr/sbin/` alleviates the issue.

PR #308, opened 2020-02-25
Lazy unmount efs volume

fenxiong:
### Summary

Lazy unmount efs volume. In case of awsvpc network mode, when we unmount the volume, task network namespace has been deleted and nfs server is no longer reachable, so umount will hang. Hence doing lazy unmount in the plugin.

### Implementation details

Add "-f" flag for umount command.

### Testing

Manually built rpm and tested with efs agent code [here](https://github.com/aws/amazon-ecs-agent/pull/2372). Verified that the volume can be cleaned up when doing lazy unmount, but cannot be cleaned up with unmount directly.

N/A

### Licensing

### DIFF
--- a/ecs-init/volumes/efs_mount_helper.go
+++ b/ecs-init/volumes/efs_mount_helper.go
@@ -92,7 +92,9 @@ func getPath(binary string) (string, error) {
 var runUnmount = runUnmountCommand
 
 func runUnmountCommand(path string, target string) error {
-	umountCmd := exec.Command(path, target)
+	// In case of awsvpc network mode, when we unmount the volume, task network namespace has been deleted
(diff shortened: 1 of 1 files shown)


sharanyad:
how long does unmount hang without force? does agent/docker have a timeout?

sharanyad:
have we tried lazy umount with -l ? 
I'm worried if using force by default will corrupt filesystems.

fenxiong:
> how long does unmount hang without force? does agent/docker have a timeout?

agent doesn't have timeout. i don't think docker has either. it basically hangs forever. i will try lazy unmount

fenxiong:
Tested and lazy unmount works as well. i updated it to do lazy unmount.

PR #309, opened 2020-02-27
Changed changelog timestamp output to UTC

sparrc:
### Summary

Outputting the time format in local time is problematic because it will change based on the operator's current timezone. Even if they are in Seattle this can change depending on the time of the year (PDT/PST).

This PR changes the output timestamps for all of the changelogs to be in UTC.

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ecs-init.spec
+++ b/packaging/amazon-linux-ami/ecs-init.spec
@@ -255,7 +255,7 @@ fi
 %endif
 
 %changelog
-* Fri Feb 21 2020 Jessie Young <youngli@amazon.com> - 1.37.0-2
+* Sat Feb 22 2020 Jessie Young <youngli@amazon.com> - 1.37.0-2
 - Cache Agent version 1.37.0
 - Add '/etc/alternatives' to mounts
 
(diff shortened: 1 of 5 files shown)


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -399,16 +399,16 @@ fi
...
 
-* Fri Oct 06 2017 Justin Haynes <jushay@amazon.com> - 1.15.0-1
+* Sat Oct 07 2017 Justin Haynes <jushay@amazon.com> - 1.15.0-1
 - Update ECS Agent version
 
-* Fri Sep 29 2017 Justin Haynes <jushay@amazon.com> - 1.14.5-1
+* Sat Sep 30 2017 Justin Haynes <jushay@amazon.com> - 1.14.5-1
 - Update ECS Agent version
 
 * Tue Aug 22 2017 Justin Haynes <jushay@amazon.com> - 1.14.4-1
```
sharanyad: any reason these below ones remain unchanged?

sparrc: These are generated from a source timestamp in https://github.com/aws/amazon-ecs-init/blob/master/scripts/changelog/CHANGELOG_MASTER

Because this spec file doesn't have the timezone, the date would only change if the source timestamp was after 4pm PST, because anything after 4pm PST would roll-over into the next day in UTC.

PR #311, opened 2020-03-12
Remove volume from state when unmount failed with not mounted

fenxiong:
### Summary

Remove volume from state when unmount failed with not mounted.

Currently when the plugin tries to unmount a volume that isn't mounted, it fails and the volume stays in the state forever. An example case that this can happen is to reboot the instance when an EFS task is running. The EFS volume is unmounted after reboot and the volume can't be removed.

Fixing by treating unmount failed with "not mounted" error as success and let the plugin remove the volume from state in that case.

### Implementation details

See above.

### Testing

Unit test added. Manually built rpm and tested that this fixed the issue.

N/A

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -15,10 +15,14 @@ package volumes
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/cihub/seelog"
 )
+
+// TODO: might be a good idea to check the mount point with a tool like mountpoint instead of matching error message.
+const notMountedErrMsg = "not mounted"
 
 // ECSVolumeDriver holds mount helper and methods for different Volume Mounts
 type ECSVolumeDriver struct {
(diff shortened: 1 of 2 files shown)


sharanyad:
is it possible to use errors from here https://github.com/aws/aws-sdk-go/blob/master/service/efs/errors.go if available?

fenxiong:
> is it possible to use errors from here https://github.com/aws/aws-sdk-go/blob/master/service/efs/errors.go if available?

I think umount call doesn't go to either mount helper or the sdk, but is calling the os directly. so don't think we can use anything here?

ecs-init/volumes/ecs_volume_driver.go
```diff
@@ -15,11 +15,14 @@ package volumes
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/cihub/seelog"
 )
 
+const notMountedErrMsg = "not mounted"
```
petderek: it might be a good idea to use a tool like mountpoint in the future instead of matching on "not mounted". Maybe add a todo?

fenxiong: yeah, added.

PR #312, opened 2020-03-12
Clean up mount path when mount failed

fenxiong:
### Summary

Currently, when mounting a volume failed for whatever reason, the mount path is left on the host. This PR adds the logic to cleanup the mount path in that case.

### Implementation details

See above.

### Testing

Modified existing unit test. Manually tested that with the fix mount path is cleaned up when mount failed.

N/A

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -173,6 +173,10 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 	err = volDriver.Create(req)
 	if err != nil {
 		seelog.Errorf("Volume %s creation failure: %v", r.Name, err)
+		cErr := a.CleanupMountPath(target)
+		if cErr != nil {
+			seelog.Warnf("Failed to cleanup mount path for volume %s: %v", r.Name, cErr)
+		}
 		return err
 	}
 	seelog.Infof("Volume %s created successfully", r.Name)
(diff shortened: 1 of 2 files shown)


ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -173,6 +173,10 @@ func (a *AmazonECSVolumePlugin) Create(r *volume.CreateRequest) error {
 	err = volDriver.Create(req)
 	if err != nil {
 		seelog.Errorf("Volume %s creation failure: %v", r.Name, err)
+		cErr := a.CleanupMountPath(target)
+		if cErr != nil {
+			seelog.Errorf("Failed to cleanup mount path for volume %s: %v", r.Name, cErr)
```
sharanyad: this can be warnf

ecs-init/volumes/ecs_volume_plugin_test.go
```diff
@@ -189,11 +189,18 @@ func TestVolumeCreateFailure(t *testing.T) {
 	createMountPath = func(path string) error {
 		return nil
 	}
+
+	var removeInvoked bool
+	removeMountPath = func(path string) error {
```
sharanyad: this should be re-assigned to the original method in defer block.

ecs-init/volumes/ecs_volume_plugin.go
fenxiong: updated

ecs-init/volumes/ecs_volume_plugin_test.go
fenxiong: good catch. fixed

PR #313, opened 2020-03-17
optimization: disable ipv6 router advertisements

sparrc:
### Testing

optimization: disable ipv6 router advertisements

### Licensing


yrral86:
I believe this change is the cause of https://github.com/aws/amazon-ecs-init/issues/323

PR #315, opened 2020-03-18
Update makefile and readme

sparrc:
### Summary

gocyclo and mockgen are needed for running tests and generating mocks, so add them to the makefile

also add a blurb about getting dependencies and generating mocks in the README

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,8 @@ get-deps:
 	go get golang.org/x/tools/cover
 	go get golang.org/x/tools/cmd/cover
 	go get golang.org/x/tools/cmd/goimports
+	go get github.com/fzipp/gocyclo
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -89,6 +89,8 @@ get-deps:
 	go get golang.org/x/tools/cover
 	go get golang.org/x/tools/cmd/cover
 	go get golang.org/x/tools/cmd/goimports
+	go get github.com/fzipp/gocyclo
```
sharanyad: am pretty sure i added this https://github.com/aws/amazon-ecs-init/blob/38904632b2cb09c768b71321c7e9aaf64622740e/Makefile#L109 🤔

PR #320, opened 2020-03-25
Change CHANGELOG_MASTER to RFC3339 to fix UTC/Local issues

sparrc:
### Summary

RFC1123 seems to always parse the time in local time despite having a timezone in the timestamp, according to the go docs (https://golang.org/pkg/time):

> Note that the RFC822, RFC850, and RFC1123 formats should be applied only to local times. Applying them to UTC times will use "UTC" as the time zone abbreviation, while strictly speaking those RFCs require the use of "GMT" in that case. In general RFC1123Z should be used instead of RFC1123 for servers that insist on that format, and RFC3339 should be preferred for new protocols. RFC3339, RFC822, RFC822Z, RFC1123, and RFC1123Z are useful for formatting; when used with time.Parse they do not accept all the time formats permitted by the RFCs. The RFC3339Nano format removes trailing zeros from the seconds field and thus may not sort correctly once formatted. 

changing CHANGELOG_MASTER to RFC3339 as this is the preferred time format.

### Testing

ran on both laptop set to PDT and "cloud desktop" set to UTC and got the same results.

### Licensing

### DIFF
--- a/scripts/changelog/changelog.go
+++ b/scripts/changelog/changelog.go
@@ -44,12 +44,13 @@ func main() {
 			for i := 3; i < len(changeStrings); i++ {
 				currentChanges = append(currentChanges, changeStrings[i])
 			}
-			thisTime, err := time.Parse(time.RFC1123, changeStrings[2])
+			thisTime, err := time.Parse(time.RFC3339, changeStrings[2])
 			handleErr(err, "error parsing time")
-			thisChange := Change{changeStrings[0],
-				changeStrings[1],
-				thisTime,
-				currentChanges,
+			thisChange := Change{
+				Version:  changeStrings[0],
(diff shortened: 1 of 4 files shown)


scripts/changelog/changelog.go
```diff
@@ -201,14 +202,17 @@ func getTopLevelChangeString(allChange []Change) string {
 // update changelog files
 // removes original file, creates updated file
 func rewriteChangelog(changelogFile string, updateString string) {
-	if _, err := os.Stat(changelogFile); err == nil {
+	_, err := os.Stat(changelogFile)
+	if err != nil {
+		handleErr(err, "unable to state changelog file: "+changelogFile)
```
fierlion: unable to 'stat'

PR #330, opened 2020-06-15
Update Go version, Add staticcheck, fix code inefficiencies found

prateekchaudhry:
### Summary

Update Go version to 1.12.
Update static-check make target by adding staticcheck tool. Uses all default checks except code style check (-ST*).
Also fixes a golang inefficiency found by it.

### Implementation details

static-check make target is updated with the command to run the check

### Testing

Tested using the updated static-check make target
`make static-check`

New tests cover the changes: 

### Licensing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 sudo: false
 go:
-    - 1.9
+    - 1.12
 before_install: ./scripts/hack/symlink-gopath-travisci
 install:
     - go get golang.org/x/tools/cover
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 sudo: false
 go:
-    - 1.9
+    - 1.12
```
fierlion, line 4: looks like this is now in line with agent
https://github.com/aws/amazon-ecs-agent/blob/master/.travis.yml#L5

PR #331, opened 2020-06-22
V1.41.0 1 stage

mssrivas:
### Summary
Release 1.41.0-1

### Testing

New tests cover the changes: 

### Licensing


Makefile
```diff
@@ -29,7 +29,7 @@ static:
 	./scripts/gobuild.sh
 
 govet:
-	go vet $(shell go list ./ecs-init/...)
+	go vet ./...
```
yunhee-l, line 32: why was this changed?

scripts/changelog/changelog.go
```diff
@@ -85,7 +85,7 @@ func main() {
 		for scanner.Scan() {
 			line := scanner.Text()
 			if strings.Contains(line, "%changelog") {
-				amazonLinuxSpecBase += fmt.Sprintln("%changelog")
+				amazonLinuxSpecBase += "%changelog"
```
yunhee-l, line 88: with removal of Sprintln, should this be "%changelog\n" ?

Makefile
sparrc: this was changed because it’s not necessary and it was missing some files in ./scripts/ directory

scripts/changelog/changelog.go
sparrc: yes I think you may be correct

mssrivas: i can fix this one

PR #332, opened 2020-06-23
Add code coverage analysis

sparrc:
### Summary

This is the init repo counterpart to https://github.com/aws/amazon-ecs-agent/pull/2484

Use Go's builtin unit test coverage profiler to verify that our unit test coverage doesn't drop with time.

Currently we are at 76.7% on linux so this will be used as the current minimum.

Also put the static-check before running tests because this is a faster check than unit tests, and can catch build errors earlier.

Also cleanup travis ci yaml file to match agent repo, and remove
unnecessary 'hack' script.

### Licensing


scripts/analyze-cover-profile
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+
+MINIMUM_TEST_COVERAGE = 76.7
+
+# coverprofile.out (see Makefile for how this is generated) will look something like this:
+#
+# github.com/aws/amazon-ecs-agent/agent/wsclient/error.go:66:						Retry						0.0%
+# github.com/aws/amazon-ecs-agent/agent/wsclient/errors.go:23:						Error						0.0%
```
yunhee-l, line 8: nit: formatting?

Makefile
```diff
@@ -32,7 +32,15 @@ govet:
...
 gotest:
-	go test -count=1 -short -v -cover ./...
+	go test -count=1 -short -v -coverprofile cover.out ./...
+
+.PHONY: analyze-cover-profile
+analyze-cover-profile: cover-profile
+	./scripts/analyze-cover-profile
+
+.PHONY: cover-profile
+cover-profile:
```
petderek, line 42: nit: if you name this the actual name of the file that gets generated, you don't need to declare it phony.

petderek: oh, also this should depend on gotest

sparrc: true, I actually had it that way to begin with. Drawback though is that then it won't regenerate the file when it sees it's already there so re-analyzing changed code would require cleaning up this file. Since all our other targets are "phony" I thought this would be confusing to most people as we don't typically need a "make clean" in between make targets atm.

scripts/analyze-cover-profile
sparrc: that's how the output actually looks, they are tabs so i guess how it lines up depends on how large github's tab spacing is.

Makefile
petderek: also, `make clean` should cleanup cover.out and coverprofile.out

Sorry for the comment stream, brain is stalling.

sparrc: > oh, also this should depend on gotest

we dont want to rerun gotest when this runs though, the cover.out file is already made

PR #334, opened 2020-07-08
Drop traffic to 127.0.0.1 that isn't originated from the host

fenxiong:
### Summary
 
Add iptable rule to drop traffic to 127.0.0.1 that isn't originated from the host. This is necessitated by the fact that we set `net.ipv4.conf.all.route_localnet` to 1.

### Implementation details

Add following iptable rule when starting up:
```
iptables -I INPUT --dst 127.0.0.0/8 ! --src 127.0.0.0/8 -m conntrack ! --ctstate RELATED,ESTABLISHED,DNAT -j DROP
```
Remove the rule upon stop.
Some refactor in unit test code to reduce duplicate code and use assert/require from testify library.

### Testing

Unit tests added/updated.

Built the rpm and verified that the expected iptable rule is added after the sevice start:
```
[ec2-user@ip-xxx ~]$ sudo iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
DROP       all  -- !ip-127-0-0-0.us-west-2.compute.internal/8  ip-127-0-0-0.us-west-2.compute.internal/8  ! ctstate RELATED,ESTABLISHED,DNAT
```
and that the rule is removed when service stopped.
Ran a few agent functional tests with the rpm including the task iam role test to make sure the local traffic forwarding to task credential endpoint isn't affected.

Drop traffic to 127.0.0.1 that isn't originated from the host.

### Licensing

### DIFF
--- a/ecs-init/exec/iptables/iptables.go
+++ b/ecs-init/exec/iptables/iptables.go
@@ -15,9 +15,11 @@ package iptables
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/amazon-ecs-init/ecs-init/exec"
 	log "github.com/cihub/seelog"
+	"github.com/pkg/errors"
 )
 
 // iptablesAction enumerates different actions for the iptables command
@@ -29,10 +31,16 @@ const (
 	credentialsProxyPort          = "80"
 	localhostIpAddress            = "127.0.0.1"
 	localhostCredentialsProxyPort = "51679"
+	localhostNetwork              = "127.0.0.0/8"
 	// iptablesAppend enumerates the 'append' action
 	iptablesAppend iptablesAction = "-A"
+	// iptablesInsert enumerates the 'insert' action
+	iptablesInsert iptablesAction = "-I"
 	// iptablesDelete enumerates the 'delete' action
 	iptablesDelete iptablesAction = "-D"
+
+	iptablesTableFilter = "filter"
+	iptablesTableNat    = "nat"
 )
 
 // NetfilterRoute implements the engine.credentialsProxyRoute interface by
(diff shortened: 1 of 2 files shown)


ecs-init/exec/iptables/iptables.go
```diff
@@ -29,10 +31,16 @@ const (
 	credentialsProxyPort          = "80"
 	localhostIpAddress            = "127.0.0.1"
 	localhostCredentialsProxyPort = "51679"
+	localhostNetwork              = "127.0.0.0/8"
 	// iptablesAppend enumerates the 'append' action
 	iptablesAppend iptablesAction = "-A"
+	// iptablesAppend enumerates the 'insert' action
```
fierlion, line 37: nit: should this comment refer to  `iptablesInsert`?

ecs-init/exec/iptables/iptables.go
```diff
@@ -61,38 +69,62 @@ func NewNetfilterRoute(cmdExec exec.Exec) (*NetfilterRoute, error) {
...
 
 // Remove removes the route for the credentials endpoint from the netfilter
 // table
 func (route *NetfilterRoute) Remove() error {
-	preroutingErr := route.modifyNetfilterEntry(iptablesDelete, getPreroutingChainArgs)
+	preroutingErr := route.modifyNetfilterEntry(iptablesTableNat, iptablesDelete, getPreroutingChainArgs)
 	if preroutingErr != nil {
 		// Add more context for error in modifying the prerouting chain
-		preroutingErr = fmt.Errorf("Error removing prerouting chain entry: %v", preroutingErr)
+		preroutingErr = fmt.Errorf("error removing prerouting chain entry: %v", preroutingErr)
```
fierlion, line 91: nit `Error removing ...` <- this happens a couple more times below.  Keep the capitalization consistent between error messages.

ecs-init/exec/iptables/iptables.go
fenxiong: good catch, fixed

ecs-init/exec/iptables/iptables.go
fenxiong: i think the error messages in this file are all lowercases i think? error level logs staring with cap is fine imo

PR #337, opened 2020-07-22
Configure Agent container logging driver

tamor1:
### Summary

Adds environment variables ECS_LOG_DRIVER and ECS_LOG_OPTS to allow a logging driver to be set for the Agent container when using ecs-init.

### Implementation details

Changes how the godocker LogConfig is set by overriding the defaults with the values from the environment variables, if they are set. The option to set max-size and max-file for the json-file logging driver with ECS_INIT_DOCKER_LOG_FILE_SIZE and ECS_INIT_DOCKER_LOG_FILE_NUM still exists, but those values will be overridden if ECS_LOG_OPTS is set. 

### Testing

Unit tests for testing the configuration, will also do smoke test.

### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -14,10 +14,14 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"runtime"
 	"strings"
+
+	"github.com/cihub/seelog"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	godocker "github.com/fsouza/go-dockerclient"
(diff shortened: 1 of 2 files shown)


ecs-init/config/common.go
```diff
@@ -202,23 +214,41 @@ func HostPKIDirPath() string {
 // AgentDockerLogDriverConfiguration returns a LogConfig object
 // suitable for used with the managed container.
 func AgentDockerLogDriverConfiguration() godocker.LogConfig {
-	maxSize := dockerJSONLogMaxSize
-	if fromEnv := os.Getenv(dockerJSONLogMaxSizeEnvVar); fromEnv != "" {
-		maxSize = fromEnv
+	driver := "json-file"
```
yunhee-l, line 217: why not add this to the list of consts?

ecs-init/config/common.go
```diff
@@ -202,23 +217,57 @@ func HostPKIDirPath() string {
...
+	exists := struct{}{}
+	validDrivers := map[string]struct{}{
+		"awslogs":    exists,
+		"fluentd":    exists,
+		"gelf":       exists,
+		"json-file":  exists,
+		"journald":   exists,
+		"logentries": exists,
+		"syslog":     exists,
+		"splunk":     exists,
```
mythri-garaga, line 229: do you think it would be better if we initialize this variable outside of this function?

Issue #336, opened 2020-07-20
Containers with multiple listening ports in 'host' networking unable to send traffic to other local ports


edhull:
### Summary
Since https://github.com/aws/amazon-ecs-init/commit/655db4e248c74a72343b6926ac482c5a2c04bc85 it is now not possible for ECS containers using 'host' networking with multiple listening ports to access other ports in the same container.

### Description
We have a single container running ClamAV and Postfix. Postfix forwards mails to Clamav (port 10026) and once finished, ClamAV sends them back to Postfix on port 10025. The processes locate each other as '127.0.0.1' from within the container.  The container is also running with 'host' networking.

Since this change was pushed, services are no longer able to send traffic to each other (connection time out) as the new INPUT deny rule is triggering. 

Flushing the INPUT chain works until amazon-ecs-init restarts and the chain is repopulated.

Attempting to insert new rules to the INPUT iptables chain is also not an option, as it appears that the amazon-ecs-init always flushing the chain as part of the 'pre-start' flag. There also does not appear to be a flag which can be passed to disable the INPUT chain from being flushed/modified. If it would be possible to provide a flag to disable this functionality, or not enable it by default, it would be very appreciated!

### Expected Behavior
Containers with 'host' networking able to access localhost services on multiple ports from within the container.

### Observed Behavior
Containers with 'host' networking listening on multiple ports are unable to send traffic locally.

### Environment Details
Docker 19.03.6-ce
Kernel: 4.14.181-142.260.amzn2.x86_64
amazon-ecs-init: 1.41.1-2

petderek:
Thanks for the detailed information! 

Would a flag to disable this behavior be sufficient for your use case? We can look at adding that, but we will probably want to leave this new behavior in as the default.

edhull:
That would absolutely be sufficient :) 

I can completely understand the methodology of default-deny and completely support it, but without an avenue available to override (or even customise) the INPUT chain the only other workaround would be an architectural refactor. Many thanks for the prompt reply!

PR #338, opened 2020-07-23
Add a flag to allow skipping local traffic filtering

fenxiong:
### Summary

Address https://github.com/aws/amazon-ecs-init/issues/336. Per https://github.com/aws/containers-roadmap/issues/976 for security purpose, an iptable rule is added to drop localhost traffic whose source address isn't in the localhost network and doesn't belong to an established connection/DNAT. As this can break some existing usage, adding a flag `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER` to allow skipping local traffic filtering as a workaround.

On Amazon Linux 1, the flag can be turned on via adding `env ECS_SKIP_LOCALHOST_TRAFFIC_FILTER=true` to `/etc/init/ecs.conf`.
On Amazon Linux 2, the flag can be turned on via adding `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER=true` to `/etc/ecs/ecs.config`.

### Implementation details

Parse env ECS_SKIP_LOCALHOST_TRAFFIC_FILTER. If set to true, skip adding/removing the relevant iptable rule.

### Testing

Unit tests added. Manually built an rpm and verified the rule is not added when the env is set to true.

Add a flag `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER` to allow skipping local traffic filtering https://github.com/aws/amazon-ecs-init/pull/338.

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,12 @@ configurations can be used to configure other properties of the ECS Agent contai
 |:----------------|:----------------------------|:------------|:-----------------------|
 | `ECS_AGENT_LABELS` | `{"test.label.1":"value1","test.label.2":"value2"}` | The labels to add to the ECS Agent container. | |
 
+Additionally, the following environment variable(s) can be used to configure the behavior of the RPM:
+
+| Environment Variable Name | Example Value(s)            | Description | Default value |
+|:----------------|:----------------------------|:------------|:-----------------------|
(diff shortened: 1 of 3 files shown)


petderek:
LGTM -- maybe update the readme too.

fenxiong:
> LGTM -- maybe update the readme too.

yeah, updated the readme.

PR #347, opened 2020-08-27
[WIP] Setup ubuntu-focal packaging

tamor1:
### Summary

Sets up ubuntu-focal directory with files to create a debian package of ecs-init for ubuntu-focal. When testing the package with MACIS, some of the EFS tests still fail, and so does the test UpdateContainerAgent, though this was expected because it uses update commands not available on Ubuntu. The EFS test contains 8 parts, usually 6 of them have been passing - it consistently fails at least TestEFS/test-efs-iam-awsvpc and TestEFS/test-efs-rw-awsvpc.

### Testing

The package is being tested by building and installing the package on an Ubuntu instance, creating an AMI from that instance, then passing that AMI into a MACIS test run, after some changes are made to MACIS setup for running with Ubuntu. 

### Licensing


packaging/ubuntu-focal/debian/amazon-ecs-init.amazon-ecs-volume-plugin.service
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
petderek, line 1: Are these identical?  Can we use symlinks instead of copying?

packaging/ubuntu-focal/debian/rules
```diff
@@ -0,0 +1,22 @@
...
+# Uncomment this to turn on verbose mode.
+export DH_VERBOSE=1
+
+# build with PIE hardening flag https://wiki.archlinux.org/index.php/Arch_package_guidelines/Security
+export GOFLAGS=-buildmode=pie
+
+%:
+	dh $@
+
+override_dh_installsystemd:
```
petderek, line 13: Does the unit file for the socket need to be listed here too?

ecs-init/docker/docker.go
```diff
@@ -23,7 +23,6 @@ import (
 
 	"github.com/aws/amazon-ecs-init/ecs-init/backoff"
 	"github.com/aws/amazon-ecs-init/ecs-init/config"
-	"github.com/aws/amazon-ecs-init/ecs-init/gpu"
```
petderek, line 26: Just an idea: If this change is going to be here for a while, we could create a .patch file and put it in the focal packaging. That way, this change is only applied when we build Ubuntu.

packaging/ubuntu-focal/debian/amazon-ecs-init.amazon-ecs-volume-plugin.service
tamor1: They are mostly identical but the path in the execstart command is slightly different than in the AL version

packaging/ubuntu-focal/debian/rules
tamor1: I'm not completely sure, the socket is supposed to be named the same as its corresponding service file and this section tells it to install amazon-ecs-init.amazon-ecs-volume-plugin.service as amazon-ecs-volume-plugin.service. I think that would be enough for them to work together and for now it seems to activating correctly with the socket, but I will look in to whether the socket file name needs to change to fall under this too.

sparrc:
Thank you @tamor1 for the great reference! I'm going to close this for now in preference of https://github.com/aws/amazon-ecs-init/pull/390

building debian packages for ecs-init is currently evolving, so we may revisit platform-specific builds at a later date.

PR #354, opened 2020-09-21
Block offhost access to agent's introspection port by default

fenxiong:
### Summary

Block offhost access to agent's introspection port by default.

During start, add iptable rule "iptables -A INPUT -i eth0 -p tcp --dport 51678 -j DROP". This blocks:
(b.1) offhost access from same subnet over instance private ip: http://$(instance_private_ip):51678;
(b.2) any access over instance public ip: http://$(instance_public_ip):51678;
(b.3) any access from container in awsvpc network mode over instance private ip: http://$(instance_public_ip):51678

This allows:
(a.1) on instance access over instance private ip: http://$(instance_private_ip):51678
(a.2) on instance access over localhost: http://localhost:51678
(a.3) same instance, from container in host network mode, over localhost and private ip: http://localhost:51678 and http://$(instance_private_ip):51678;
(a.4) same instance, from container in bridge network mode, over private ip: http://$(instance_private_ip):51678

This behavior is configurable via env ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS (when set to true, the above mentioned rule is not added). To set this, for AL1, add `env ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS=true` to /etc/init/ecs.conf; for AL2, add `ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS=true` to /etc/ecs/ecs.config.

### Implementation details

Added logic in iptable.go to add/drop iptables rule `iptables -A INPUT -i eth0 -p tcp --dport 51678 -j DROP`

### Testing

Added unit test. Built the rpm and verified the rule is added by default, and not added when opted out.

Block offhost access to agent's introspection port by default. Configurable via env `ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS`.

### Licensing

### DIFF
--- a/ecs-init/exec/iptables/iptables.go
+++ b/ecs-init/exec/iptables/iptables.go
@@ -43,6 +43,9 @@ const (
 
 	iptablesTableFilter = "filter"
 	iptablesTableNat    = "nat"
+
+	offhostIntrospectionAccessConfigEnv = "ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS"
+	agentIntrospectionServerPort        = "51678"
 )
 
 // NetfilterRoute implements the engine.credentialsProxyRoute interface by
@@ -77,9 +80,16 @@ func (route *NetfilterRoute) Create() error {
 	}
 
 	if !skipLocalhostTrafficFilter() {
-		err = route.modifyNetfilterEntry(iptablesTableFilter, iptablesInsert, getInputChainArgs)
+		err = route.modifyNetfilterEntry(iptablesTableFilter, iptablesInsert, getLocalhostTrafficFilterInputChainArgs)
 		if err != nil {
 			return err
+		}
+	}
+
+	if !allowOffhostIntrospection() {
+		err = route.modifyNetfilterEntry(iptablesTableFilter, iptablesInsert, getBlockIntrospectionOffhostAccessInputChainArgs)
+		if err != nil {
+			log.Errorf("Error adding input chain entry to block offhost introspection access: %v", err)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ecs-init/exec/iptables/iptables.go
```diff
@@ -77,12 +80,19 @@ func (route *NetfilterRoute) Create() error {
...
+		err = route.modifyNetfilterEntry(iptablesTableFilter, iptablesInsert, getLocalhostTrafficFilterInputChainArgs)
 		if err != nil {
 			return err
 		}
 	}
 
+	if !allowOffhostIntrospection() {
+		err = route.modifyNetfilterEntry(iptablesTableFilter, iptablesInsert, getBlockIntrospectionOffhostAccessInputChainArgs)
+		if err != nil {
+			log.Errorf("Error adding input chain entry to block offhost introspection access: %v", err)
```
angelcar, line 92: Should we return the error here? 
Just logging it means to let the previous behavior go, right?

ecs-init/exec/iptables/iptables.go
```diff
@@ -95,11 +105,18 @@ func (route *NetfilterRoute) Remove() error {
 		preroutingErr = fmt.Errorf("error removing prerouting chain entry: %v", preroutingErr)
 	}
 
-	var inputErr error
+	var localhostInputError, introspectionInputError error
```
angelcar, line 108: nit: move `introspectionInputError` closer to its scope in line 115

ecs-init/exec/iptables/iptables.go
fenxiong: my intention is to make this "best effort" so only log the error and continue here instead of failing. one situation i can think of is that customer can rename "eth0" to something else. in that case it doesn't seem necessary for us to block them

ecs-init/exec/iptables/iptables.go
fenxiong: these two are used in line 129

angelcar (approved):

petderek (approved):
lgtm

PR #363, opened 2020-10-15
Add bind mount for agent

cyastella:
### Summary

This is part of the change for ecs-agent to detect if an EC2 instance is `exec-enabled`, in other words, if the instance has the `exec` capability. Related change in [amazon-ecs-agent](https://github.com/aws/amazon-ecs-agent):
(add link to pr in ecs-agent)

Update ecs-init to add bind mounts to the ecs-agent containter for exec-for-ECS capability detection. For an EC2 instance to be considered as `exec-enabed`, the instance needs to have these resources required by SSM:

* SSM agent binaries
* ~~SSM agent configuration files~~ (not validated for now)
* certs files for SSM agent

### Implementation details

For now, we only want to make sure all resources required by SSM (binaries, certs) exist. Since capabilities are stamped by ecs-agent, we need to provide a mechanism for the ecs-agent container to inspect the file system on host instance. To do this, we add bind mounts from the host instance to the ecs-agent container to check if all the following files exist:

```
/host/path/to/ecs-exec-dep
|-- bin
    |-- amazon-ssm-agent
    |-- ssm-session-worker

/host/path/to/certs
|-- tls-ca-bundle.pem
```

these get bind mounted to the ecs-agent container as:

```
/capabilities/exec/
|-- bin
    |-- amazon-ssm-agent
    |-- ssm-session-worker
|-- certs
    |-- tls-ca-bundle.pem
```

One detail is to only create a bind mount if the source file/directory exists on host, we do **not** want to create an empty directory (with `sudo`) on host instance with docker. If the require file/directory does not exist on the instance, we want the ecs-agent to know that these resources do not exist, so the `exec` capability will not be added.

### Testing

Manual test:

1. `sudo systemctl start ecs`
2. when all dependencies exist, inspect `amazon-ecs-agent:latest` container to check if all corresponding binds in `HostConfig` and mounts are added
3. remove one of the dependencies, `sudo systemctl start ecs`
4. inspect `amazon-ecs-agent:latest` container to check the corresponding binds and mounts are no longer added

* `TestGetCapabilityExecBinds`
* `TestPathExistsAndValid`
* `TestIsDir`
* `TestIsFile`

add bind mounts from SSM dependency (files/directories) on host to ecs-agent container for exec capability detection 

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -125,6 +125,8 @@ var pluginDirs = []string{
 	pluginSpecFilesUsrDir,
 }
 
+var isPathValid = defaultIsPathValid
+
 // Client enables business logic for running the Agent inside Docker
 type Client struct {
 	docker dockerclient
@@ -406,7 +408,7 @@ func (c *Client) getHostConfig(envVarsFromFiles map[string]string) *godocker.Hos
 	binds = append(binds, getDockerPluginDirBinds()...)
 
 	// only add bind mounts when the src file/directory exists on host; otherwise docker API create an empty directory on host
-	binds = append(binds, getCapabilityExecBinds(pathExistsAndValid)...)
+	binds = append(binds, getCapabilityExecBinds()...)
 
 	return createHostConfig(binds)
 }
@@ -442,8 +444,7 @@ func getDockerPluginDirBinds() []string {
 	return pluginBinds
 }
 
-// take pathPredicate as an argument for unit testing
-func getCapabilityExecBinds(pathPredicate func(string, func(os.FileInfo) bool) (bool, error)) []string {
+func getCapabilityExecBinds() []string {
 	hostResourcesDir := filepath.Join(hostCapabilitiesResourcesRootDir, capabilityExecName)
 	containerResourcesDir := filepath.Join(containerCapabilitiesResourcesRootDir, capabilityExecName)
 
@@ -452,37 +453,27 @@ func getCapabilityExecBinds(pathPredicate func(string, func(os.FileInfo) bool) (
 	// bind mount the entire /host/dependency/path/exec/bin folder for higher flexibility
 	// minimal change required to add other ssm binaries as dependency in the future (just need to be placed inside the bin directory)
 	hostBinDir := filepath.Join(hostResourcesDir, capabilityExecHostBinRelativePath)
-	if exists, err := pathPredicate(hostBinDir, isDir); err == nil && exists {
+	if isPathValid(hostBinDir, true) {
 		binds = append(binds, hostBinDir+":"+filepath.Join(containerResourcesDir, capabilityExecContainerBinRelativePath)+readOnly)
 	}
 
 	// bind mount this specific cert file for now, CertsDir and CertsFile might be changed to be configurable in the future
 	hostCert := filepath.Join(capabilityExecHostCertsDir, capabilityExecRequiredCert)
-	if exists, err := pathPredicate(hostCert, isFile); err == nil && exists {
+	if isPathValid(hostCert, false) {
 		binds = append(binds, hostCert+":"+filepath.Join(containerResourcesDir, capabilityExecContainerCertsRelativePath, capabilityExecRequiredCert)+readOnly)
 	}
 
 	return binds
 }
 
-func pathExistsAndValid(path string, predicate func(os.FileInfo) bool) (bool, error) {
+func defaultIsPathValid(path string, shouldBeDirectory bool) bool {
 	fileInfo, err := os.Stat(path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
-		}
-		return false, err
+		return false
 	}
 
-	return predicate(fileInfo), nil
-}
-
-func isDir(fileInfo os.FileInfo) bool {
-	return fileInfo != nil && fileInfo.IsDir()
-}
-
-func isFile(fileInfo os.FileInfo) bool {
-	return fileInfo != nil && !fileInfo.IsDir()
+	isDirectory := fileInfo.IsDir()
+	return (isDirectory && shouldBeDirectory) || (!isDirectory && !shouldBeDirectory)
 }
 
 // nvidiaGPUDevicesPresent checks if nvidia GPU devices are present in the instance
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -107,6 +107,16 @@ const (
 	iptablesUsrLibDir   = "/usr/lib"
 	iptablesLib64Dir    = "/lib64"
 	iptablesUsrLib64Dir = "/usr/lib64"
+
+	hostCapabilitiesResourcesRootDir      = "/home/ec2-user"
+	containerCapabilitiesResourcesRootDir = "/capabilities"
+
+	capabilityExecName                       = "exec"
+	capabilityExecHostBinRelativePath        = "ssm-agent/linux_amd64"
+	capabilityExecHostCertsDir               = "/etc/pki/ca-trust/extracted/pem"
+	capabilityExecHostCertsFilename          = "tls-ca-bundle.pem"
+	capabilityExecContainerBinRelativePath   = "bin"
+	capabilityExecContainerCertsRelativePath = "certs"
 )
 
 var pluginDirs = []string{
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker.go
```diff
@@ -427,6 +432,65 @@ func getDockerPluginDirBinds() []string {
 	return pluginBinds
 }
 
+// take pathPredicate as an argument for unit testing
+func getCapabilityExecBinds(pathPredicate func(path string, predicate func(fileInfo os.FileInfo) bool) (bool, error)) []string {
+	const (
+		capabilityExecName = "exec"
+
+		hostCapabilitiesResourcesRootDir      = "/home/ec2-user/capabilities-deps"
```
sharanyad, line 440: this seems different from what we used in agent code. We don't need a separate dir for capabilities deps, since it would be a duplicate of the binaries we already add in the AMIs.

sharanyad, line 440: nit: check lint

ecs-init/docker/docker.go
```diff
@@ -427,6 +432,65 @@ func getDockerPluginDirBinds() []string {
...
+	// bind mount this specific cert file for now, CertsDir and CertsFile might be changed to be configurable in the future
+	hostCertsFile := filepath.Join(hostCertsDir, hostCertsFilename)
+	if exists, err := pathPredicate(hostCertsFile, isFile); err == nil && exists {
+		binds = append(binds, hostCertsFile+":"+filepath.Join(containerCapabilityExecResourcesDir, containerCertsRelativePath, hostCertsFilename)+readOnly)
+	}
+
+	return binds
+}
+
+func pathExistsAndValid(path string, predicate func(fileInfo os.FileInfo) bool) (bool, error) {
```
sharanyad, line 481: could you add unit tests for this method?

ecs-init/docker/docker.go
```diff
@@ -427,6 +432,65 @@ func getDockerPluginDirBinds() []string {
...
+	}
+	isFile := func(fileInfo os.FileInfo) bool {
+		return !fileInfo.IsDir()
+	}
+	hostCapabilityExecResourcesDir := filepath.Join(hostCapabilitiesResourcesRootDir, capabilityExecName)
+	containerCapabilityExecResourcesDir := filepath.Join(containerCapabilitiesResourcesRootDir, capabilityExecName)
+
+	binds := []string{}
+
+	// bind mount the entire /deps/exec/bin folder for higher flexibility
```
sharanyad, line 465: the path mentioned here and `hostCapabilitiesResourcesRootDir` are different. Could you update the comment?

ecs-init/docker/docker.go
```diff
@@ -427,6 +432,65 @@ func getDockerPluginDirBinds() []string {
 	return pluginBinds
 }
 
+// take pathPredicate as an argument for unit testing
+func getCapabilityExecBinds(pathPredicate func(path string, predicate func(fileInfo os.FileInfo) bool) (bool, error)) []string {
+	const (
```
sharanyad, line 437: if these constants are moved out of this method, they can be used in unit tests as well :)

ecs-init/docker/docker.go
cyastella: didn't notice that we already defined `/home/ec2-user/ssm-agent/linux_amd64` earlier, thanks for pointing it out!

cyastella: do you mean `go fmt`? thought this is properly formatted already

ecs-init/docker/docker.go
cyastella: I would argue that there is not much value doing that since this is a very thin wrapper on top of `os.Stat`, adding UTs for this would be almost identical to testing golang's `os` library

ecs-init/docker/docker.go
cyastella: this comment is intended to use an abstracted path since we might later make this configurable; yet good point, I'll update the path to `/host/path/dependency/exec/bin` so it's more obvious in terms of what it means

ecs-init/docker/docker.go
cyastella: good point, was trying to create some encapsulation so it's easier to read, will move those out

ecs-init/docker/docker.go
cyastella: side note on this, I think what we should do is to have a directory for capabilities' dependencies (maybe `ecs-capabilities-deps`, etc.?); otherwise, dropping `ssm-agent` on the instance's home path seems out of nowhere

ecs-init/docker/docker.go
angelcar: +1 on adding test, it's easy to return the wrong thing, like returning `false` when it should have been `true`. If someone changes this code in the future and returns the wrong thing, then exec might break and we would realize late in the process, better to catch things early

ecs-init/docker/docker.go
```diff
@@ -393,6 +404,10 @@ func (c *Client) getHostConfig(envVarsFromFiles map[string]string) *godocker.Hos
 	}
 
 	binds = append(binds, getDockerPluginDirBinds()...)
+
+	// only add bind mounts when the src file/directory exists on host; otherwise docker API create an empty directory on host
+	binds = append(binds, getCapabilityExecBinds(pathExistsAndValid)...)
```
angelcar, line 409: is there a unit test that needs to consider this?

ecs-init/docker/docker.go
angelcar: The paths used in the agent code (i.e. /home/ec2-user/ssm-agent/linux_amd64) are arbitrary and were temporary for testing. I think this is where we should define the ideal path where the resources should live. 
How are we going to place these resources into the ami? I would prefer to use one of the directories we're using for toher things like the boltdb. 

I propose:
`/var/lib/ecs/capabilities/exec`

What do you think?

ecs-init/docker/docker.go
cyastella: got it, thanks for the explanation! will add unit test to verify that we're using the expected api

ecs-init/docker/docker.go
```diff
@@ -427,6 +442,48 @@ func getDockerPluginDirBinds() []string {
...
+	isDir := func(fileInfo os.FileInfo) bool {
+		return fileInfo.IsDir()
+	}
+	isFile := func(fileInfo os.FileInfo) bool {
+		return !fileInfo.IsDir()
+	}
+	hostResourcesDir := filepath.Join(hostCapabilitiesResourcesRootDir, capabilityExecName)
+	containerResourcesDir := filepath.Join(containerCapabilitiesResourcesRootDir, capabilityExecName)
+
+	binds := []string{}
```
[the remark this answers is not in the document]
cyastella: thanks for pointing it out! I thought this would not work:

```go
var a []string = nil
b := []string{}
b = append(b, a...)
```

tested and it worked, will change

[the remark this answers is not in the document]
angelcar: you don't need the explicit `var a []string = nil`... just `var a []string` will do since the default value is `nil`

ecs-init/docker/docker.go
sharanyad: why not just `/var/lib/ecs/exec` , why do we need capabilities?

ecs-init/docker/docker.go
sharanyad: yeah, we should add on to this https://github.com/aws/amazon-ecs-init/blob/b5d2758a0ba113cd06d5e9867e15c49c206d3207/ecs-init/docker/docker_test.go#L32

cyastella: don't think it'd be worthwhile to only test against `append()`, so I'll update the current `TestGetCapabilityExecBinds` test a bit to include this behavior

angelcar: I didn't meat to test that append works, I mean to update the relevant tests for the function you're adding behavior to. As Sharanya points out, some tests are expecting 19 bind mounts, this means that probably we're expecting 20 now, under certain conditions... that's what we should test

cyastella: got it!

ecs-init/docker/docker_test.go
```diff
@@ -247,6 +251,20 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
...
+	hostCapabilityExecResourcesDir := filepath.Join(hostCapabilitiesResourcesRootDir, capabilityExecName)
+	bindSourcePaths := []struct {
+		path      string
+		predicate func(os.FileInfo) bool
+	}{
+		{filepath.Join(hostCapabilityExecResourcesDir, capabilityExecHostBinRelativePath), isDir},
+		{filepath.Join(capabilityExecHostCertsDir, capabilityExecRequiredCert), isFile},
+	}
+	for _, path := range bindSourcePaths {
+		if existsAndValid, err := pathExistsAndValid(path.path, path.predicate); err == nil && existsAndValid {
```
cyastella, line 264: not handling `err` here since `pathExistsAndValid` is already called in `docker.go`, errors should have been thrown there

ecs-init/docker/docker.go
cyastella: checked offline, resolving =)

ecs-init/docker/docker_test.go
```diff
@@ -248,6 +251,20 @@ func validateCommonCreateContainerOptions(opts godocker.CreateContainerOptions,
...
```
angelcar, line 264: so, if I'm understanding correctly, for this unit test we depend on the actual files to exist in the host? 
The unit test should be mocking out external dependencies and just testing behavior. 

e.g.
* given that the cert file exists and is valid (mocked out), assert that cert-related binds were added

* given that the cert file does NOT exist, or is valid (mocked out), assert that cert-related binds were not added

ecs-init/docker/docker.go
```diff
@@ -427,6 +442,49 @@ func getDockerPluginDirBinds() []string {
...
+// take pathPredicate as an argument for unit testing
+func getCapabilityExecBinds(pathPredicate func(string, func(os.FileInfo) bool) (bool, error)) []string {
+	hostResourcesDir := filepath.Join(hostCapabilitiesResourcesRootDir, capabilityExecName)
+	containerResourcesDir := filepath.Join(containerCapabilitiesResourcesRootDir, capabilityExecName)
+
+	var binds []string
+
+	// bind mount the entire /host/dependency/path/exec/bin folder for higher flexibility
+	// minimal change required to add other ssm binaries as dependency in the future (just need to be placed inside the bin directory)
+	hostBinDir := filepath.Join(hostResourcesDir, capabilityExecHostBinRelativePath)
```
angelcar, line 454: I think we should do individual bind mounts. If the bin directory exists but is empty, this code would say we're all good, when we're not.

ecs-init/docker/docker_test.go
angelcar: As things stand, it's hard to mock things because the real `pathExistsAndValid` is passed in `getHostConfig`. We might need to change the implementation strategy here, this is getting a bit more complicated than necessary

cyastella: good point! I am looking at these tests (that call `validateCommonCreateContainerOptions()`) as integration tests so I did not any mocks, but you are definitely right that this is by itself also hard to mock the predicate function, would love to know the alternatives

ecs-init/docker/docker.go
cyastella: current approach (mounting `/bin`, `/certs`, etc.) is to keep all the capability detection logic on `ecs-agent` side, so it would be more flexible on the `ecs-init` side: all the actual binaries can be verified on the `ecs-agent` side (where it is also easier to share constants for binaries names, etc.). individual bind mounts, `ecs-init` should still always say we're good (launch `ecs-agent` container) since having ssm-required resources is optional (even if we don't, it shouldn't fail the ecs service)

this way, even if we update the validation logic (e.g. require more binaries), users only need to update `ecs-agent`. and `ecs-init` won't need to be added. actually similar idea for `/certs`, but I think we would always require only 1 cert file so I limited it to only 1 (the filename can also be moved to `ecs-agent` part as well)

angelcar: I see, you're right. ecs agent will have this responsibility

ecs-init/docker/docker_test.go
```diff
@@ -751,3 +755,158 @@ func TestGetDockerSocketBind(t *testing.T) {
...
+		isPathValid = defaultIsPathValid
+	}()
+
+	mockFS := NewMockfileSystem(mockCtrl)
+	mockDocker := NewMockdockerclient(mockCtrl)
+
+	mockFS.EXPECT().ReadFile(config.InstanceConfigFile()).Return(nil, errors.New("not found")).AnyTimes()
+	mockFS.EXPECT().ReadFile(config.AgentConfigFile()).Return(nil, errors.New("not found")).AnyTimes()
+	mockDocker.EXPECT().CreateContainer(gomock.Any()).Do(func(opts godocker.CreateContainerOptions) {
+		validateCommonCreateContainerOptions(opts, t)
```
angelcar, line 803: We're validating the expected number of bind mounts, which is good. But can we also verify in here that the actual contents are ok? Let's do this check after calling `validateCommonCreateContainerOptions`, as opposed to doing the checks inside `validateCommonCreateContainerOptions`

angelcar (commented):
Much cleaner = )

angelcar (approved):

sharanyad (approved):
thanks for the extensive tests!

PR #366, opened 2020-10-22
update capability name, add config to agent binds

cyastella:
### Summary

this change updates bind mounts to `ecs-agent` from `ecs-init` to use the constants defined in https://github.com/aws/amazon-ecs-agent/pull/2687.

### Implementation details

update capability name to match the latest decision spec. add config folder as an a agent bind to allow ecs-agent to write (ssm configuration files) to host file system

### Testing

manual test include starting the `ecs-agent` container and inspecting, automated unit test is also added

existing tests updated:

* TestStartAgentWithExecBinds
* TestGetCapabilityExecBinds

update capability ecs-exec name, add config to agent binds

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -109,16 +109,14 @@ const (
 	iptablesUsrLib64Dir = "/usr/lib64"
 
 	hostCapabilitiesResourcesRootDir      = "/var/lib/ecs/deps"
-	containerCapabilitiesResourcesRootDir = "/capabilities"
+	containerCapabilitiesResourcesRootDir = "/managed-agents"
 
-	capabilityExecName                        = "execute-command"
-	capabilityExecHostBinRelativePath         = "bin"
-	capabilityExecContainerBinRelativePath    = "bin"
-	capabilityExecHostConfigRelativePath      = "config"
-	capabilityExecContainerConfigRelativePath = "config"
-	capabilityExecHostCertsDir                = "/etc/pki/ca-trust/extracted/pem"
-	capabilityExecContainerCertsRelativePath  = "certs"
-	capabilityExecRequiredCert                = "tls-ca-bundle.pem"
+	capabilityExecName               = "execute-command"
+	capabilityExecBinRelativePath    = "bin"
+	capabilityExecConfigRelativePath = "config"
+	capabilityExecCertsRelativePath  = "certs"
+	capabilityExecHostCertsDir       = "/etc/pki/ca-trust/extracted/pem"
+	capabilityExecRequiredCert       = "tls-ca-bundle.pem"
 
 	execAgentLogRelativePath = "/exec"
 )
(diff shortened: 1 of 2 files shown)


ecs-init/docker/docker_test.go
```diff
@@ -804,34 +824,7 @@ func TestStartAgentWithExecBinds(t *testing.T) {
...
-			for _, bind := range agentBinds {
-				if bind == expectedBind {
-					added = true
-					break
-				}
-			}
-
-			assert.True(t, added, "%v should be added to HostConfig.Binds", expectedBind)
-		}
+		assert.Subset(t, opts.HostConfig.Binds, expectedExecBinds)
```
angelcar, line 827: nice one!

ecs-init/docker/docker.go
```diff
@@ -108,15 +108,17 @@ const (
...
 
-	capabilityExecName                       = "exec"
-	capabilityExecHostBinRelativePath        = "ssm-agent/linux_amd64"
-	capabilityExecContainerBinRelativePath   = "bin"
-	capabilityExecHostCertsDir               = "/etc/pki/ca-trust/extracted/pem"
-	capabilityExecContainerCertsRelativePath = "certs"
-	capabilityExecRequiredCert               = "tls-ca-bundle.pem"
+	capabilityExecName                        = "execute-command"
+	capabilityExecHostBinRelativePath         = "bin"
+	capabilityExecContainerBinRelativePath    = "bin"
```
angelcar, line 116: hmm, I don't think we will ever want to diverge from this... not opposed to ship this tho, but don't think it's necessary to have two `"bin"` and two `"config"`

ecs-init/docker/docker.go
```diff
@@ -108,15 +108,17 @@ const (
...
+	capabilityExecHostConfigRelativePath      = "config"
+	capabilityExecContainerConfigRelativePath = "config"
+	capabilityExecHostCertsDir                = "/etc/pki/ca-trust/extracted/pem"
```
sharanyad, line 119: this will change too, if we are copying instance certs to exec deps folder.

ecs-init/docker/docker.go
```diff
@@ -453,17 +455,26 @@ func getCapabilityExecBinds() []string {
...
 	if isPathValid(hostBinDir, true) {
-		binds = append(binds, hostBinDir+":"+filepath.Join(containerResourcesDir, capabilityExecContainerBinRelativePath)+readOnly)
+		binds = append(binds,
+			hostBinDir+":"+filepath.Join(containerResourcesDir, capabilityExecContainerBinRelativePath)+readOnly)
 	}
 
-	// bind mount this specific cert file for now, CertsDir and CertsFile might be changed to be configurable in the future
+	// bind mount the entire /host/dependency/path/execute-command/config folder
+	// in read-write mode to allow ecs-agent to write config files to host file system
+	hostConfigDir := filepath.Join(hostResourcesDir, capabilityExecHostConfigRelativePath)
```
sharanyad, line 467: why is config folder needed for capability detection? Initially, we wont have any config in the AMI, so this check feels unnecessary

angelcar: it needs to be bind mounted into the acs agent container in read/write mode

ecs-init/docker/docker_test.go
cyastella: yay!

ecs-init/docker/docker.go
cyastella: was not sure about this either, thanks for pointing it out

ecs-init/docker/docker.go
cyastella: right, we want to make sure that there is 1 place for the agent to write files to

ecs-init/docker/docker.go
cyastella: thanks for pointing it out! discussed offline, will keep this instance cert path for now; we'll change later once we change to the new model (copying instance certs to the exec deps folder)

angelcar (approved):

ecs-init/docker/docker.go
```diff
@@ -109,16 +109,14 @@ const (
 	iptablesUsrLib64Dir = "/usr/lib64"
 
 	hostCapabilitiesResourcesRootDir      = "/var/lib/ecs/deps"
-	containerCapabilitiesResourcesRootDir = "/capabilities"
+	containerCapabilitiesResourcesRootDir = "/managed-agents"
```
sharanyad, line 112: Don't think we need the 'capability' word in these constants anymore

cyastella: good point, I'm still keeping the `capability` prefix before `exec` since `exec` is one capability, e.g. `capabilityExecName`

sharanyad (approved):

PR #378, opened 2020-12-01
Go mod tool dependencies

mythri-garaga:
### Summary

This PR includes changes to use Go Modules to manage tool dependencies and updates Makefile targets accordingly.

### Implementation details

- Added `./ecs-init/tools.go` to list all the tool dependencies using an import statement.
- Updated makefile target `get-deps` to install tool dependencies from `./ecs-init/tools.go`
- Added `./ecs-init/Makefile` with the following targets and updated `./Makefile` accordingly
	- generate
	- govet
	- gocyclo
	- test
	- importcheck
	- goimports
	- static-check
- Updated file path for  `coverprofile.out` in  `scripts/analyze-cover-profile` 
- Use go1.15 instead of go1.9 in  `scripts/dockerfiles/test.dockerfile` as Go Modules available after Go version 1.1
- Added static-check in  `scripts/test`
- File changes in `vendor` directory after running `go mod vendor`

### Testing

Locally tested all the Makefile targets.

New tests cover the changes: 

### Licensing


shubham2892 (approved):

ecs-init/Makefile
```diff
@@ -0,0 +1,38 @@
+.PHONY: generate
```
fierlion, line 1: Are you breaking these out into a separate makefile for clarity/readability?  Can you explain why not just expand the top level Makefile?
this is a departure from the ecs-agent: https://github.com/aws/amazon-ecs-agent/tree/master/agent
is the plan to create a similar agent level Makefile?

mythri-garaga: We have our main module in `./ecs-init` (go.mod file) which is now the module path and go command can only be invoked from module path hence the move of these targets to `./ecs-init`

We are doing similar for ecs-agent too - https://github.com/aws/amazon-ecs-agent/commit/bba546eb55a57a4b2c4cbcfdc9897bfc270d3731

fierlion (approved):

PR #386, opened 2021-01-20
Prevent restart on exit code 5

shubham2892:
### Summary
Init is currently configured to restart on any failure. Today what happens is --

- Agent exits with code 5
- ecs-init, who monitors agent process status, received this error and kills itself with exit code -1.
- ecs-init restarts due to the restart policy 
- ecs-init restarts agent

Agent should not restart when it exits with exit code 5 as it is a terminal error code. This PR fixes this behavior. 

The behavior will be changed to --  
- Agent exits with code 5
- ecs-init, who monitors agent process status, received this error and kill itself with exit code 5.
- ecs-init does not restart as the restart policy is modified to not restart itself on exit code 5

### Testing
Agent exiting with exit code 5 can be replicated by adding an arbitrary capability to ecs.config - 
```
ECS_INSTANCE_ATTRIBUTES={"custom attribute !@#$%^&*": "custom_attribute_value"}
```

Montoring ecs-init after adding this capability, I can see Agent restarting again and again -- 

```
2021-01-21T19:50:36Z [INFO] Removing existing agent container ID: 2724ace14f79ebbb5f32c4b5614f85b36885377ee975b3378f287796c2b639d8
2021-01-21T19:50:36Z [INFO] Starting Amazon Elastic Container Service Agent
2021-01-21T19:50:38Z [INFO] Agent exited with code 5
2021-01-21T19:50:38Z [ERROR] agent exited with terminal exit code
2021-01-21T19:50:38Z [INFO] stop
2021-01-21T19:50:38Z [INFO] Stopping Amazon Elastic Container Service Agent
2021-01-21T19:50:38Z [INFO] Container name: /ecs-agent
2021-01-21T19:50:38Z [INFO] Agent is already stopped
2021-01-21T19:50:38Z [INFO] post-stop
2021-01-21T19:50:38Z [INFO] Cleaning up the credentials endpoint setup for Amazon Elastic Container Service Agent
2021-01-21T19:50:48Z [INFO] pre-start
2021-01-21T19:50:48Z [INFO] start
2021-01-21T19:50:48Z [INFO] Container name: /ecs-agent
2021-01-21T19:50:48Z [INFO] Removing existing agent container ID: 8582cbf6ebe0af6ff066139a22dea311a9c1c6995b6a5921a5bb593482ed2fe0
2021-01-21T19:50:48Z [INFO] Starting Amazon Elastic Container Service Agent



2021-01-21T19:50:49Z [INFO] Agent exited with code 5
[...]

PR #389, opened 2021-01-21
Add generic rpm packaging

fenxiong:
### Summary

Add generic rpm packaging (one that is not specific to platform). 

### Implementation details

The changes include the following parts:
1. `packaging/generic-rpm/ecs-init.spec`: the rpm spec file used to built the rpm. This is written based on what we have on AL https://github.com/fenxiong/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs-init.spec, with the following notable differences:
* Remove all the AL1/upstart related logic and only use the AL2 part;
* Omit the distro part in https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs-init.spec#L36;
* Omit the docker dependency https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs-init.spec#L60. This is mainly because we have less control over docker outside of AL. it can come from "docker-ce", "docker" or "docker-ee", or even not from rpm. so i omit this dependency in packaging;
* Omit the provides statement https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs-init.spec#L71-L160. they are already outdated and not maintained so probably not needed i think. we can add them later if we find use case for them;
* Omit the changelog section since the code will not be in main branch in a while. they can be added later when ready to merge to main branch.
* Omit the volume plugin for now. They can be added later if needed. currently I have not tested the plugin.
2. `packaging/generic-rpm/ecs.service`: the systemd service file. this is basically the same as the one for AL, except i added "/var/lib/ecs/ecs.config" as another optional environment file, which is where we put "ECS_EXTERNAL" env and allows ecs init to use it in https://github.com/aws/amazon-ecs-init/pull/383
3. `Makefile`: added a target generic-rpm to build the rpm.

### Testing

 Built the rpm and tested it on:
* Centos 7
* Centos 8
* RHEL 7
* Fedora 32
* opensuse/Tumbleweed

Verified that the service runs fine on these platforms.

N/A

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,18 @@ BUILDROOT/ecs-agent.tar:
 	mkdir -p BUILDROOT
 	curl -o BUILDROOT/ecs-agent.tar ${AGENT_URL}
 
+.generic-rpm-done:
+	./scripts/update-version.sh
+	cp packaging/generic-rpm/amazon-ecs-init.spec amazon-ecs-init.spec
+	cp packaging/generic-rpm/ecs.service ecs.service
+	tar -czf ./sources.tgz ecs-init scripts
+	test -e SOURCES || ln -s . SOURCES
+	rpmbuild --define "%_topdir $(PWD)" -bb amazon-ecs-init.spec
+	find RPMS/ -type f -exec cp {} . \;
+	touch .rpm-done
+
+generic-rpm: .generic-rpm-done
+
 .PHONY: deb
 deb: .deb-done
 .deb-done: BUILDROOT/ecs-agent.tar
@@ -156,7 +168,7 @@ clean:
 	-rm -rf ./BUILDROOT BUILD RPMS SRPMS SOURCES SPECS
 	-rm -rf ./x86_64
 	-rm -f ./amazon-ecs-init_${VERSION}*
-	-rm -f .srpm-done .rpm-done
+	-rm -f .srpm-done .rpm-done .generic-rpm-done
 	-rm -f .deb-done
 	-rm -f cover.out
 	-rm -f coverprofile.out
--- a/packaging/generic-rpm/amazon-ecs-init.spec
+++ b/packaging/generic-rpm/amazon-ecs-init.spec
@@ -1,0 +1,96 @@
+# Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the
+# "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+%global gobuild_tag generic-rpm
+%global _cachedir %{_localstatedir}/cache
+%global bundled_agent_version %{version}
+%global no_exec_perm 644
+
+%ifarch x86_64
+%global agent_image %{SOURCE3}
+%endif
+%ifarch aarch64
+%global agent_image %{SOURCE4}
+%endif
+
+Name:           amazon-ecs-init
+Version:        1.48.1
+Release:        1
+License:        Apache 2.0
+Summary:        Amazon Elastic Container Service initialization application
+ExclusiveArch:  x86_64 aarch64
+
+Source0:        sources.tgz
+Source1:        ecs.conf
+Source2:        ecs.service
+# x86_64 Container agent docker image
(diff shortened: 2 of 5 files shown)


packaging/generic-rpm/ecs.service
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
+
+[Service]
```
mssrivas, line 22: Lets include shugy's commit here - #386

packaging/generic-rpm/ecs-init.spec
```diff
@@ -0,0 +1,99 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+%global gobuild_tag al2
```
mssrivas, line 15: Should this tag still be al2?

sparrc: i suppose we should change the tag, and we should fix the existing build tag divergence. Since this is needed for the deb package too maybe we can follow up with a separate PR

Makefile
```diff
@@ -113,6 +113,18 @@ srpm: .srpm-done
 
 rpm: .rpm-done
 
+.generic-rpm-done:
```
sparrc, line 116: should this also depend on the sources.tgz target like .rpm-done does?

packaging/generic-rpm/ecs-init.spec
```diff
@@ -0,0 +1,99 @@
...
+%post
+# Symlink the bundled ECS Agent at loadable path.
+ln -sf %{basename:%{agent_image}} %{_cachedir}/ecs/ecs-agent.tar
+%systemd_post ecs
+
+%postun
+%systemd_postun
+
+%posttrans
+systemctl enable ecs
```
sparrc, line 97: remove this now to match current deb package (sorry for the misdirection earlier)

packaging/generic-rpm/ecs.service
fenxiong: yeah, added

packaging/generic-rpm/ecs-init.spec
fenxiong: unfortunately if the tag is unspecified, it defaults to what al1 wants and uses a cgroup location that only works for al1 https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/config/config_unspecified.go#L19. so if we want to not specify build tag here, we need to add a build tag for al1 and make sure that works. i added a comment here about it and we can follow up later

Makefile
fenxiong: sources.tgz is only copying files under AL folder https://github.com/aws/amazon-ecs-init/blob/master/Makefile#L87-L92. so i added some similar source preparation commands in the .generic-rpm-done target: https://github.com/aws/amazon-ecs-init/pull/389/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R118-R120

packaging/generic-rpm/ecs-init.spec
fenxiong: yeah, done

packaging/generic-rpm/amazon-ecs-init.spec
```diff
@@ -0,0 +1,99 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO: The build tag is specified to al2 because if unspecified it defaults to what al1 expects, but
+# the al2 config is the one that specifies the correct cgroup for all the rpm based system. Should probably
+# refactor the code to fix this logic.
+%global gobuild_tag al2
```
mssrivas, line 18: Can we change this now since Cam made the change for ubuntu tag? Or it is still not feasible?

fenxiong: no build tag defaults to al1 logic, so we still need to avoid it. i added a separate config file for generic rpm https://github.com/aws/amazon-ecs-init/pull/389/files#diff-27b431da0991cf90dec63480aafc5e13631b908f3ed44df8c7e9485dc16c6b24R1, to avoid abusing the al2 tag

mssrivas (approved):

sparrc (approved):

PR #390, opened 2021-01-28
Add generic deb packaging

sparrc:
## Summary

Add generic deb packaging (one that is not specific to platform).

## Implementation details

Some changes from the ubuntu-trusty packaging:

- changed maintainer from skarp@amazon to ecs-agent-dev@amazon
- changed deb changelog distribution from "trusty" to "stable" and added `--lintian-opts --suppress-tags bad-distribution-in-changes-file` to suppress "bad distribution" linter error. This is because "stable" is not a valid distribution name for Ubuntu.
- added a list of possible docker engine sources instead of only docker-engine: `docker-ce (>= 17.12.0) | docker-engine (>= 1.6.0) | docker-ee | docker.io`
- ~~added a post-install script to enable the ecs service~~
- added `override_dh_auto_install` rule to add the bundled agent container tarball to the cache directory
- added ecs.service file, removed ecs.conf (only supporting systemd for now)
- remove DEB_SIGN option as it is not used
- add `arm64` as a valid architecture to the control file
- download and cache the relevant agent tarball version for the tagged version of init. Previously we were not caching this so ecs would have just downloaded the latest version on startup.
- install systemd service file via `dh_installsystemd --no-start --no-enable --name=ecs`

## Testing

Built the deb and tested it on:

- [x] ubuntu 20.04
- [x] ubuntu 18.04
- [x] debian 10
- [x] debian 9
- [x] ubuntu 20.04 ARM
- [x] ubuntu 18.04 ARM
- [x] debian 10 ARM
- [x] debian 9 ARM

Verified that the service runs fine on these platforms.

## Description for the changelog

N/A

## Licensing


nmeyerhans:
> Omit the docker dependency. This is mainly because we have less control over docker outside of AL. it can come from "docker-ce", "docker" or "docker-ee", or even not from deb. so i omit this dependency in packaging;

The debian packaging format allows optional dependencies. You should specify a Docker dependency using something like:
`Depends: docker-ce | docker-ee | docker.io | docker`

This obviously won't cover the case where it's not installed via a package, but for people going that route, they will have other issues, and [other ways of satisfying the dependency](https://packages.debian.org/buster/equivs).

We should not be leaving critical dependencies unspecified

packaging/generic-deb/debian/dirs
```diff
@@ -1,3 +1,4 @@
 etc/ecs
 var/log/ecs
 var/lib/ecs
+var/cache/ecs
```
mssrivas, line 4: is the data file also suppose to be included?

packaging/generic-deb/ecs.service
```diff
@@ -0,0 +1,33 @@
...
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://aws.amazon.com/documentation/ecs/
+Requires=docker.service
+After=docker.service
+
+[Service]
```
mssrivas, line 21: We should also include the new 'RestartPreventExitStatus' from the bug fix Shugy did - https://github.com/aws/amazon-ecs-init/pull/386

packaging/generic-deb/debian/dirs
sparrc: yes i will add

Makefile
```diff
@@ -113,16 +113,27 @@ srpm: .srpm-done
...
+ifeq (${ARCH},x86_64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-v${VERSION}.tar
+else ifeq (${ARCH},aarch64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-arm64-v${VERSION}.tar
+endif
+
+download-agent:
+	curl -o /tmp/ecs-agent-v${VERSION}.tar ${AGENT_URL}
+
+deb: download-agent
```
nmeyerhans, line 126: You should use a stamp file for dependencies, similar to what's done with the `.rpm-done` file. e.g.
```
.PHONY: deb
deb: .deb-done
.deb-done:
  <rest of recipe>
```

Makefile
```diff
@@ -113,16 +113,27 @@ srpm: .srpm-done
...
-	cp packaging/ubuntu-trusty/ecs.conf ecs.conf
-	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init ecs.conf scripts README.md
+ARCH:=$(shell uname -m)
+ifeq (${ARCH},x86_64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-v${VERSION}.tar
+else ifeq (${ARCH},aarch64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-arm64-v${VERSION}.tar
+endif
+
+download-agent:
```
nmeyerhans, line 123: This recipe should be declared PHONY, or better yet its name should match that of the downloaded image so it can be used properly as a dependency of downstream targets. If you get the dependencies done right, and want (for example) to build a .rpm and a .deb, make will know that it doesn't need to retrieve the agent multiple times.

Makefile
```diff
@@ -113,16 +113,27 @@ srpm: .srpm-done
...
 	mkdir -p BUILDROOT
-	cp -r packaging/ubuntu-trusty/debian BUILDROOT/debian
+	cp -r packaging/generic-deb/debian BUILDROOT/debian
 	cp -r ecs-init BUILDROOT
-	cp packaging/ubuntu-trusty/ecs.conf BUILDROOT
+	cp packaging/generic-deb/ecs.service BUILDROOT
 	cp -r scripts BUILDROOT
 	cp README.md BUILDROOT
-	cd BUILDROOT && debuild $(shell [ "$(DEB_SIGN)" -ne "0" ] || echo "-uc -us")
+	cd BUILDROOT && debuild $(shell [ "$(DEB_SIGN)" -ne "0" ] || echo "-uc -us") --lintian-opts --suppress-tags bad-distribution-in-changes-file
```
nmeyerhans, line 136: It'd probably be cleaner to store the "-uc -us" options in a variable, conditional on DEB_SIGN being set, and reference that variable here instead of invoking the shell. Maybe store the lintian flags in a variable, too. Something like
```
debuild_flags:=--lintian-opts --suppress-tags bad-distribution-in-changes-file
ifeq (${DEB_SIGN},0)
debuild_flags+= -uc -us
endif
...
deb:
        ...
        cd BUILDROOT && debuild ${debuild_flags}
        ...

File: packaging/generic-deb/debian/README.source [hunk 6]
```diff
@@ -0,0 +1,14 @@
+Steps to build and install package:
+
+1. Install build dependencies
+    sudo apt-get update -y && sudo apt-get install -y devscripts build-essential lintian git curl golang-go debhelper
+
+2. Build the package by running
+    make clean deb
+
+3. Install the package with
+    sudo dpkg -i amazon-ecs-init_1.45.0-1_amd64.deb
```
nmeyerhans, line 10: Modern versions of apt will handle local files, and since apt resolves dependencies, it'd be preferable to use

     sudo apt install ./amazon-ecs-init_1.45.0-1_amd64.deb

here. Note that the leading `./` is important.

packaging/generic-deb/debian/README.source
```diff
@@ -0,0 +1,14 @@
...
+
+4. Install docker and start ecs service
+    sudo apt-get install docker.io -y # or via docker repos: https://docs.docker.com/engine/install/
```
nmeyerhans, line 13: The ecs-init package depends on a docker installation, so this should happen first. Although if you use the `apt` invocation I described above to install the ecs-init package, Docker will be installed automatically.

It's probably fine to list the docker installation as an explicit step, though, in case the user wants to customize the configuration or something like that before setting up ECS. Just put it before the ecs-init installation in the steps.

packaging/generic-deb/debian/amazon-ecs-init.install
```diff
@@ -0,0 +1,2 @@
+amazon-ecs-init usr/libexec/
```
nmeyerhans, line 1: It doesn't look like you've listed /usr/libexec in the `debian/dirs` file.

packaging/generic-deb/debian/postinst
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -e
+
+systemctl enable ecs
```
nmeyerhans, line 4: You should actually not need this file at all. If you do include it, you should not explicitly call `systemctl` in it yourself. If you do include a postinst script, you should be sure to include `#DEBHELPER#` on a line by itself, which debhelper will fill in with the appropriate `systemctl` invocations (and anything else made necessary by its integration with the package)

packaging/generic-deb/debian/rules
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+VERSION := $(shell git describe --tags | sed -e 's/v//' -e 's/-.*//')
```
nmeyerhans, line 3: Assuming that you're building in a git working copy will cause problems. The source artifacts that you're generating in the build process (the .orig.tar.* and *.dsc files) won't contain the git metadata, and thus it won't be possible to use them to generate binaries.

Since you've got version information in `debian/changelog` by the time this runs, you should use `dpkg-parsechangelog -S Version` in place of `git describe --tags`

Makefile
sparrc: I'm not aware of a use-case where we actually need DEB_SIGN, so I will just remove it for now and we can add later if it's needed for building the official deb packages that we distribute.

nmeyerhans: Yeah, I think that's reasonable. The signatures are attached to the source artifacts, not the binaries .deb files, and you're not currently doing anything with those anyway.

Makefile
sparrc: will do

Makefile
sparrc: done

packaging/generic-deb/debian/README.source
sparrc: building ecs-init doesn't require docker, so I don't think it's necessary to install before building the deb. We download the agent container from s3 so we are not building any docker containers as part of this packaging

packaging/generic-deb/debian/amazon-ecs-init.install
sparrc: done

packaging/generic-deb/debian/postinst
sparrc: I tried to remove this file and build the package but the service is installed in a "disabled" state:
```
% sudo systemctl status ecs
● ecs.service - Amazon Elastic Container Service - container agent
     Loaded: loaded (/lib/systemd/system/ecs.service; disabled; vendor preset: enabled)
     Active: inactive (dead)
       Docs: https://aws.amazon.com/documentation/ecs/
```

Is there some way to only enable the service without a postinst script?

I tried using `dh ... --with systemd` in the rules makefile, but that also auto-starts the service, which we do not want.

packaging/generic-deb/debian/rules
sparrc: changed to reading from a text file next to the agent tarball

packaging/generic-deb/debian/README.source
sparrc: done

packaging/generic-deb/debian/README.source
nmeyerhans: My comment was with respect to the installation of ecs-init, not building it. Build happens in step 1. You have ecs-init installation happening in step 3, and docker installation in step 4. You either need to re-order 3 and 4, or combine them.

packaging/generic-deb/debian/postinst
nmeyerhans: Expected behavior on Debian-derived systems is to enable and start the service upon installation. However, if you really don't want that, you can use the `--no-start` option to `dh_installsystemd`. Put the following in your `override_dh_auto_install` recipe in `debian/rules`:
```
	cp ecs.service debian/
	dh_installsystemd --name ecs --no-start
	dh_systemd_enable
```
With this in place, you should see quite a bit of content inserted into the postinst, prerm, and postrm scripts in the binary package. This will take care of all the complexity around service management across package upgrades, removals, etc.

If you're not going to start it, then I also recommend that you don't mark it as enabled, either. Use the `--no-enable` flag to `dh_installsystemd`.  Having a service installed and enabled but not actually running is rarely an expected scenario.

You should also update the `clean` recipe to remove the `debian/ecs.service` file you created there, and you can remove `ecs.service` from `amazon-ecs-init.install`

Makefile
```diff
@@ -113,16 +112,33 @@ srpm: .srpm-done
...
-	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init ecs.conf scripts README.md
+ARCH:=$(shell uname -m)
+ifeq (${ARCH},x86_64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-v${VERSION}.tar
+else ifeq (${ARCH},aarch64)
+	AGENT_URL=https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-arm64-v${VERSION}.tar
+endif
+
+/tmp/ecs-agent-version:
+	echo "${VERSION}" >/tmp/ecs-agent-version
```
nmeyerhans, line 123: Don't use a static name in a world-writable directory for this. Consider what happens if /tmp/ecs-agent-version is a symlink to some other file.

You should either invoke `mktemp` to create a file with an unpredictable name, or create this file in the build directory somewhere.

Makefile
```diff
@@ -113,16 +112,33 @@ srpm: .srpm-done
...
+
+/tmp/ecs-agent.tar: /tmp/ecs-agent-version
+	curl -o /tmp/ecs-agent.tar ${AGENT_URL}
```
nmeyerhans, line 126: Same here; don't use static filenames in /tmp

Makefile
```diff
@@ -113,16 +112,33 @@ srpm: .srpm-done
...
+.deb-done: /tmp/ecs-agent.tar
+	./scripts/update-version.sh
+	cp packaging/generic-deb/ecs.service ecs.service
+	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init ecs.service scripts README.md
 	mkdir -p BUILDROOT
-	cp -r packaging/ubuntu-trusty/debian BUILDROOT/debian
+	cp -r packaging/generic-deb/debian BUILDROOT/debian
 	cp -r ecs-init BUILDROOT
-	cp packaging/ubuntu-trusty/ecs.conf BUILDROOT
+	cp packaging/generic-deb/ecs.service BUILDROOT
```
nmeyerhans, line 137: You should copy to `BUILDROOT/debian`, where `dh_installsystemd` will pick it up. I described this more in another comment.

nmeyerhans (changes requested):
This is looking good. The lack of `dh_installsystemd` and the unsafe use of `/tmp` are the major remaining issues.

packaging/generic-deb/debian/rules
```diff
@@ -0,0 +1,21 @@
...
+# -*- makefile -*-
+VERSION := $(shell cat /tmp/ecs-agent-version)
+
+# Uncomment this to turn on verbose mode.
+export DH_VERBOSE=1
+
+%:
+	dh $@
+
+build:
```
nmeyerhans, line 11: This target should be named `override_dh_auto_build` rather than just `build`

packaging/generic-deb/ecs.service
```diff
@@ -0,0 +1,33 @@
+# Copyright 2014-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
nmeyerhans, line 1: In some other comments I suggested moving this file around during packaging, but instead you should probably just move it under the `debian/` subdirectory in git.

sparrc: done

packaging/generic-deb/debian/rules
sparrc: done

Makefile
sparrc: done

Makefile
sparrc: done

Makefile
sparrc: I've changed this to put ecs-agent.tar in the BUILDROOT directory.

ecs-agent-version, as it's not a binary file, I put it in ../BUILDROOT. I was not able to put it in BUILDROOT as dpkg did not like that it was an unmanaged/changed file.

packaging/generic-deb/debian/postinst
sparrc: this is done.

We are enabling the service because we want it to autostart when the instance is restarted. We don't want it to start, however, because the user needs to be able to setup their config file before starting the service. As in, the normal installation process would look like:

```
apt install -y ./amazon-ecs-init.deb
echo "ECS_CLUSTER=myCluster" > /etc/ecs/ecs.config
systemctl start ecs
```

packaging/generic-deb/debian/rules
sparrc: using dpkg-parsechangelog now

packaging/generic-deb/debian/postinst
nmeyerhans: Instead of `systemctl start ecs`, you could recommend `systemctl enable --now ecs`. This lets you install the unit file in a stopped and disabled state, which is more consistent with the workflow you expect. Leaving it enabled means that it'll start if the instance is rebooted for some reason, even if the user hasn't finished configuring it.

sparrc: that's a good point, will make the change

packaging/generic-deb/debian/README.source
sparrc: done

packaging/generic-deb/ecs.service
sparrc: yes we can add that once we've merged in his fix for the exit code handling too

nmeyerhans (approved):
```
        _    _    _
     __|_|__|_|__|_|_______
   _|_     SHIP IT!       _|__
  |o o o o o o o o o o o o o /
~'`~'`~'`~'`~'`~'`~'`~'`~'`~'`~
```
This looks great.

packaging/generic-deb/debian/control
```diff
@@ -1,16 +1,16 @@
...
 Vcs-Git: git://github.com/aws/amazon-ecs-init.git
 Vcs-Browser: http://github.com/aws/amazon-ecs-init
 
 Package: amazon-ecs-init
-Architecture: amd64
-Depends: ${shlibs:Depends}, ${misc:Depends}, docker-engine (>= 1.6.0)
+Architecture: amd64 arm64
+Depends: ${shlibs:Depends}, ${misc:Depends}, systemd, docker-ce (>= 17.12.0) | docker-engine (>= 1.6.0) | docker-ee | docker.io
 Description: Starts the Amazon ECS Agent
  amazon-ecs-init may be run to register an EC2 instance as an Amazon ECS
```
mssrivas, line 15: Change that?

mssrivas (approved):

fenxiong (approved):

PR #395, opened 2021-03-09
packaging/generic-deb: use agent version from file instead of tag

fenxiong:
### Summary

When we build a new version of ecs init package, the new agent version is available from file ecs-init/ECSVERSION, but not available from tag because the tag is created when we publish the release, but we build the package before that. This causes deb package build to use an agent version older than the release version.

Therefore, use the agent version from the ECSVERSION file instead of from `git describe --tags`.

### Implementation details

Get agent version from `cat ecs-init/ECSVERSION` instead of from `git describe --tags`.

### Testing

Built the deb package successfully. Verified that the version is read from file:
```
root@ip-172-31-12-102:/var/lib/ec2-stage/go/src/github.com/aws/amazon-ecs-init# make deb
./scripts/update-version.sh
++ dirname ./scripts/update-version.sh
+ CURRENTDIR=./scripts
++ cat ./scripts/../ecs-init/ECSVERSION
+ version=1.50.0
...
```

N/A

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@
 # CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and
 # limitations under the License.
-VERSION := $(shell git describe --tags | sed -e 's/v//' -e 's/-.*//')
+VERSION = $(shell cat ecs-init/ECSVERSION)
 
 .PHONY: dev generate lint static test build-mock-images sources rpm srpm govet
 


sparrc (approved):

Makefile
```diff
@@ -11,7 +11,8 @@
 # CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and
 # limitations under the License.
-VERSION := $(shell git describe --tags | sed -e 's/v//' -e 's/-.*//')
+ROOT := $(shell pwd)
+VERSION := $(shell cat $(ROOT)/ecs-init/ECSVERSION)
```
nmeyerhans, line 15: setting and using `ROOT` in this way is redundant. Just do:
`VERSION := $(shell cat ecs-init/ECSVERSION)`

You probably also want `=` instead of `:=`

fenxiong: yeah, updated

nmeyerhans (approved):
LGTM.

BTW, newer versions of make support reading data directly from files, so you may be able to eliminate the shell/cat invocation at some point. I wouldn't count on that for now though.

mssrivas (approved):

PR #398, opened 2021-03-16
Install script

Realmonia:
### Summary

* Some changes from newer releases
* Copy ecs-init install script
* Further changes for ecs-init install script including:
 * GPG check for SSM agent
 * GPG check for Init
 * Show license info at the end of installation
 * Verify agent is working at the end of installation

### Testing

It passed with the ubuntu 20.04 vagrant vm box setup by following the setup guide in https://w.amazon.com/bin/view/EC2/Project_Madison/Orchestrator/Projects/RunAnywhere/Testing

### Licensing


scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+    fail
+fi
+
+S3_BUCKET="amazon-ecs-agent-packages-preview"
+RPM_PKG_NAME="amazon-ecs-init-$ECS_VERSION.$ARCH.rpm"
+DEB_PKG_NAME="amazon-ecs-init-$ECS_VERSION.$ARCH_ALT.deb"
+
+# TODO [Update before release] Change urls to be region specific buckets
+if [ -z "$RPM_URL" ]; then
+    RPM_URL="https://ecs-init-packages-testing.s3.amazonaws.com/amazon-ecs-init-latest.$ARCH_ALT.rpm"
```
fenxiong, line 135: this should change to "https://$S3_BUCKET.s3.amazonaws.com/$RPM_PKG_NAME" (and change S3_BUCKET to ecs-init-packages-testing). this way we just need to change the bucket at the end and it allows us to test `--ecs-version` option. same for the deb url

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+    register-ssm-agent
+
+    ok
+}
+
+ssm-agent-signature-verify() {
+    echo "Verifying the signature of SSM Agent"
+
+    # TODO [Update before release] Change this url to main repo master branch
+    curl -o "$dir/amazon-ssm-agent.gpg" "https://github.com/Realmonia/amazon-ecs-init/blob/ssmGpg/scripts/amazon-ssm-agent.gpg"
```
fenxiong, line 283: do you not need to use raw content for this? i.e. `https://raw.githubusercontent.com/Realmonia/amazon-ecs-init/ssmGpg/scripts/amazon-ssm-agent.gpg`

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+    gpg --import "$dir/amazon-ssm-agent.gpg" &> "$dir/tempOut"
```
fenxiong, line 284: i think we should check the fingerprint before importing it. this can be done by the following:
(1) get fingerprint: `local fr=$(gpg --quiet --with-colons --with-fingerprint amazon-ssm-agent.gpg | awk -F: '$1 == "fpr" {print $10;}')`
(2) check fingerprint value, exit if unmatched
(3) import it: `gpg --import amazon-ssm-agent.gpg`

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+
+    if grep -Fxq "BAD signature" "$dir/verifyOut"; then
+        echo "SSM Agent GPG verification fail. Stop the installation of the SSM Agent. Please contact AWS Support."
+        fail
+    else
+        echo "SSM Agent GPG verification passed. Install the SSM Agent."
+    fi
+    rm "$dir/verifyOut"
+
+    ok
```
fenxiong, line 315: ok is only needed when you had a "try" before. maybe change line 280 to `try "Verifying the signature of SSM Agent"`

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+            register-ssm-agent
+            return
+        fi
+    else
+        local dir
+        dir="$(mktemp -d)"
+        ssm-agent-signature-verify
+        case "$PKG_MANAGER" in
+        dnf)
+            dnf install -y "https://s3.$REGION.amazonaws.com/amazon-ssm-$REGION/latest/linux_$ARCH_ALT/amazon-ssm-agent.rpm"
```
fenxiong, line 259: since ssm-agent-signature-verify downloads the rpm already, we should avoid duplicate download and just install the local one, same for yum

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+        if [ ! "$curlResult" == "null" ] && [ -n "$curlResult" ]; then
+            success=1
+            break
+        fi
+    done
+    if [ $success == 0 ]; then
+        echo "ECS Agent verification failed. Please check logs at /var/log/ecs/ecs-agent.log"
+        fail
+    else
+        echo "ping ECS Agent registered successfully!"
```
fenxiong, line 507: nit: since we got the container instance arn, maybe just print it here for people to see for convenience

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+
+    if grep -Fxq "BAD signature" "$dir/verifyOut"; then
+        echo "ECS Init GPG verification fail. Stop the installation of the ECS Agent. Please contact AWS Support."
+        fail
+    else
+        echo "ECS Init GPG verification passed. Install the ECS Agent."
+    fi
+    rm "$dir/verifyOut"
+
+    ok
```
fenxiong, line 487: same comment as https://github.com/aws/amazon-ecs-init/pull/398/files#r596248713

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+        echo "ECS_BACKEND_HOST=$ECS_ENDPOINT" >>/var/lib/ecs/ecs.config
+    fi
+    systemctl enable ecs
+    systemctl start ecs
+
+    ok
+}
+
+ecs-init-signature-verify() {
+    echo "Verifying the signature of ECS Init Artifact"
```
fenxiong, line 471: nit for clarity: "ECS Init Artifact" -> "amazon-ecs-init package"

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+        rm -rf "$dir"
+    fi
+    # register the instance
+    register-ssm-agent
+
+    ok
+}
+
+ssm-agent-signature-verify() {
+    echo "Verifying the signature of SSM Agent"
```
fenxiong, line 280: nit for clarity: "SSM Agent" -> "amazon-ssm-agent package"

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+        gpg --verify "$dir/ssm-agent.rpm.sig" "$dir/ssm-agent.rpm" &> "$dir/verifyOut"
+        ;;
+    apt)
+        curl -o "$dir/ssm-agent.deb" "https://s3.$REGION.amazonaws.com/amazon-ssm-$REGION/latest/debian_$ARCH_ALT/amazon-ssm-agent.deb"
+        curl -o "$dir/ssm-agent.deb.sig" "https://s3.$REGION.amazonaws.com/amazon-ssm-$REGION/latest/debian_$ARCH_ALT/amazon-ssm-agent.deb.sig"
+        gpg --verify "$dir/ssm-agent.deb.sig" "$dir/ssm-agent.deb" &> "$dir/verifyOut"
+        ;;
+    esac
+
+    if grep -Fxq "BAD signature" "$dir/verifyOut"; then
```
sparrc, line 307: instead of relying on the output going to a file and grepping for an error, I think it would be better to just check if the `gpg` command errors directly.

ie, this if-statement could look something like (probably changing the name of the deb/rpm file to a variable so we only have to have it once):

```
if gpg --verify "$dir/ssm-agent.deb.sig" "$dir/ssm-agent.deb"; then
    echo SUCCESS
else
    echo UH OH
fi
```

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,526 @@
...
+ecs-init-signature-verify() {
+    echo "Verifying the signature of ECS Init Artifact"
+
+    #TODO [Update before release] Update links here to use prod urls (or urls specified by $DEB_URL or $RPM_URL)
+    curl -o "$dir/amazon-ecs-init.gpg" "https://ecs-init-packages-testing.s3.amazonaws.com/amazon-ecs-public-key.gpg"
+    gpg --import "$dir/amazon-ecs-init.gpg"
+
+    gpg --verify "$1" "$2" &> "$dir/verifyOut"
+
+    if grep -Fxq "BAD signature" "$dir/verifyOut"; then
```
sparrc, line 479: same comment as I made for the ssm signature verification (https://github.com/aws/amazon-ecs-init/pull/398/files#r596305112), we should check for the gpg command to error directly rather than grepping for errors.

scripts/amazon-ssm-agent.gpg
```diff
@@ -0,0 +1,30 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
sparrc, line 1: where does this come from? does the SSM team not provide this in any public location?

Realmonia: According to the info here https://docs.aws.amazon.com/systems-manager/latest/userguide/verify-agent-signature.html we have to copy it..

scripts/ecs-anywhere-install.sh
Realmonia: Sorry missed that.. Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Nice catch. Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Sure thing.

Realmonia: I actually have it at above lines. I will combine the two.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Make sense and great suggestion.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
Realmonia: Updated.

scripts/ecs-anywhere-install.sh
```diff
@@ -0,0 +1,520 @@
...
+        echo "amazon-ecs-init GPG verification passed. Install amazon-ecs-init."
+    else
+        echo "amazon-ecs-init GPG verification failed. Stop the installation of amazon-ecs-init. Please contact AWS Support."
+        fail
+    fi
+
+    ok
+}
+
+show-link-to-console() {
```
fenxiong, line 480: this is not used. should remove?

fenxiong (commented):
looks good. one minor comment

scripts/ecs-anywhere-install.sh
Realmonia: Oh sorry didn't include it in the latest commit. Will remove.

fenxiong (approved):
make sure to squash your commits before merging

sparrc (approved):
lgtm

Realmonia:
Code merged.

PR #403, opened 2021-04-19
A few fixes for ecs anywhere install

fenxiong:
### Summary

* Removed ecs init rpm dependency on cloud-init.final: this is to allow the script to be ran in userdata (or by cloud-init in general);
* Fixed ssm agent install error on suse due to gpg signature: added `--no-gpg-checks` option when install. the ssm agent gpg signature is a detached signature and we verify explicitly before install.
* Added an optional flag --no-start: this is mainly for testing purpose so we can replace the agent being used for testing.

### Implementation details

Described above.

### Testing

Tested on ubuntu20, centos8 and opensuse15. Tested `--no-start` option working and ssm/docker/ecs agent can be started fine after the script with `sudo systemctl start amazon-ssm-agent` and `sudo systemctl start ecs`. Also checked that without `--no-start` the script still works fine.

N/A

### Licensing

### DIFF
--- a/scripts/ecs-anywhere-install-instruction.md
+++ b/scripts/ecs-anywhere-install-instruction.md
@@ -33,4 +33,6 @@ Usage of ./ecs-anywhere-install.sh
     	(optional) ecs endpoint that can be used to different requests to ECS internal endpoints
   --skip-registration
     	(optional) if this is enabled, ssm agent install and instance registration with ssm is skipped
+  --no-start
+    	(optional) if this flag is provided, ssm agent, docker and ecs agent will not be started by the script despite being installed.
 ```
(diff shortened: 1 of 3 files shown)


scripts/ecs-anywhere-install-instruction.md
```diff
@@ -33,4 +33,6 @@ Usage of ./ecs-anywhere-install.sh
     	(optional) ecs endpoint that can be used to different requests to ECS internal endpoints
   --skip-registration
     	(optional) if this is enabled, ssm agent install and instance registration with ssm is skipped
+  --no-start
+    	(optional) if this flag is provided, ssm agent, docker and ecs agent will not be started by the script despite being installed.
```
mssrivas, line 37: nit: We aren't stoping docker from running are we?

fenxiong: i don't think we explicitly start docker in the script. docker is started as a required dependency when we run `systemctl start ecs`. so skipping `systemctl start ecs` means skipping to start docker too 🤔

Realmonia (approved):
LGTM

sparrc (approved):

PR #409, opened 2021-05-07
Install script fixups

sparrc:
- check for invalid or empty cli flags
- support ssm agent installed via ubuntu snap
- check that ssm agent is running when skipping registration
- simplify ssm registration check
- skip gpg package verify if server does not have gpg command available

### Summary

Some ecs anywhere install script bumps found while testing.

### Testing

This was tested on:
- ubuntu
- centos
- debian
- sles

New tests cover the changes: 

NA

### Licensing

### DIFF
--- a/scripts/ecs-anywhere-install.sh
+++ b/scripts/ecs-anywhere-install.sh
@@ -27,6 +27,22 @@ if [ -f "/sys/fs/cgroup/cgroup.controllers" ]; then
     fail
 fi
 
+SSM_SERVICE_NAME="amazon-ssm-agent"
+SSM_BIN_NAME="amazon-ssm-agent"
+if [ "$(systemctl is-enabled snap.amazon-ssm-agent.amazon-ssm-agent.service)" == "enabled" ]; then
+    echo "Detected SSM agent installed via snap."
+    SSM_SERVICE_NAME="snap.amazon-ssm-agent.amazon-ssm-agent.service"
+    SSM_BIN_NAME="/snap/amazon-ssm-agent/current/amazon-ssm-agent"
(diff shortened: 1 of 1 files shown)


scripts/ecs-anywhere-install.sh
```diff
@@ -27,6 +27,22 @@ if [ -f "/sys/fs/cgroup/cgroup.controllers" ]; then
...
+SSM_BIN_NAME="amazon-ssm-agent"
+if [ "$(systemctl is-enabled snap.amazon-ssm-agent.amazon-ssm-agent.service)" == "enabled" ]; then
+    echo "Detected SSM agent installed via snap."
+    SSM_SERVICE_NAME="snap.amazon-ssm-agent.amazon-ssm-agent.service"
+    SSM_BIN_NAME="/snap/amazon-ssm-agent/current/amazon-ssm-agent"
+fi
+SSM_MANAGED_INSTANCE_ID=""
+
+check-option-value() {
+    if [ "${2:0:2}" == "--" ]; then
```
angelcar, line 40: is this meant to test that the 2nd parameter is not `--`?  If I do `./ecs-anywhere-install.sh --opt1 --opt2` I think this will complain, is that the intention?

sparrc: yes you're right, I think I put this check in a spot that is too broad. We shouldn't be checking on single args like --skip-registration, we be do want to check on args like `--cluster $CLUSTER_NAME`

sparrc: fixed

angelcar (approved):

PR #417, opened 2021-05-17
ecs-anywhere-install: allow sandboxed apt installations

otterley:
Ensure the temporary directories into which package artifacts are downloaded are world-readable, thereby allowing the unprivileged `_apt` user to open the packages and allow for a sandboxed install.

This fix eliminates warnings such as:

```
N: Download is performed unsandboxed as root as file '/tmp/tmp.ZkW0dDLEph/amazon-ecs-init-latest.amd64.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
```

### Licensing

### DIFF
--- a/scripts/ecs-anywhere-install.sh
+++ b/scripts/ecs-anywhere-install.sh
@@ -317,6 +317,7 @@ install-ssm-agent() {
             curl-helper "$dir/$SSM_DEB_PKG_NAME" "$SSM_DEB_URL"
             curl-helper "$dir/$SSM_DEB_PKG_NAME.sig" "$SSM_DEB_URL.sig"
             ssm-agent-signature-verify "$dir/$SSM_DEB_PKG_NAME.sig" "$dir/$SSM_DEB_PKG_NAME"
+            chmod -R a+rX "$dir"
             dpkg -i "$dir/ssm-agent.deb"
             ;;
         dnf | yum | zypper)
@@ -477,6 +478,7 @@ install-ecs-agent() {
             curl-helper "$dir/$DEB_PKG_NAME.asc" "$DEB_URL.asc"
             ecs-init-signature-verify "$dir/$DEB_PKG_NAME.asc" "$dir/$DEB_PKG_NAME"
         fi
+        chmod -R a+rX "$dir"
         apt install -y "$dir/$DEB_PKG_NAME"
         rm -rf "$dir"
         ;;


scripts/ecs-anywhere-install.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e # exit on failure
+umask 022
```
sparrc, line 3: is this a necessary part of this fix? If i understand correctly I would think that we only need one of `chmod` and `umask` ?

sparrc: I would generally prefer if the fix for this one file didn't have to change the permissions of every file created by this script. Could we chmod just the deb file after downloading it to `755`?

scripts/ecs-anywhere-install.sh
```diff
@@ -307,6 +308,7 @@ install-ssm-agent() {
     else
         local dir
         dir="$(mktemp -d)"
+        chmod 755 "$dir"
```
sparrc, line 311: this could be moved down to only happen just before `dpkg -i` is run so that it doesn't affect rpms unecessarily, making it recursive might remove the need for the umask change.

scripts/ecs-anywhere-install.sh
```diff
@@ -470,6 +472,7 @@ install-ecs-agent() {
 
     local dir
     dir="$(mktemp -d)"
+    chmod 755 "$dir"
```
sparrc, line 475: same comment as above, I think we can do this just before the `apt install`

scripts/ecs-anywhere-install.sh
otterley: The intent was to put some control around the environment because we don't know what the current value of the `umask` is. If it is `077` then all files created in this script will be readable/writable only by the `root` user. This has adverse consequences for the `apt install` later on because it needs to run as user `_apt`. 

If you're concerned about adverse side effects (can you think of any?) then it's not harmful to remove this and `chmod` the downloaded files instead as needed.

scripts/ecs-anywhere-install.sh
otterley: I'm indifferent - I see no harm in doing this, but if you think it's important, then we can be more precise.

otterley:
@sparrc Updated per our discussion, please re-review

sparrc (approved):

fenxiong (approved):

sparrc:
lgtm, I also confirmed that both +r and +X are necessary to get rid of this error message, thanks @otterley !

PR #418, opened 2021-05-18
install script fix: remove dependency on gpg key server.

fenxiong:
### Summary

Install script fix: remove dependency on gpg public key server. Instead of getting public key from key server, store the public key on github and use it in signature verification. This change is made mainly because it has been realized that the gpg public key server might not be particularly stable and can become unreachable from time to time (e.g. search for error like "gpg: keyserver receive failed: No name" on the internet and it can be seen that people run into issue due to the key server not available). Therefore, stop using gpg key server and instead directly download and import the public key.

### Implementation details

Store public key in the repo, download and install it for signature verification. This is similar to what is already done for ssm agent package. Also, make a change so when --rpm-url/--deb-url is specified, we still check a signature, which allow our e2e tests to verify signature. And added a separate option --skip-gpg-check just so that when we need to do manual test on a custom build we don't need to generate signature.

### Testing

Ran e2e tests on centos and ubuntu.

Install script fix: remove dependency on gpg public key server

### Licensing


Realmonia (approved):
LGTM

sparrc (approved):

PR #451, opened 2021-11-10
add exec prerequisites to ecs-anywhere installation script

lydiafilipe:
### Summary
This pr makes changes to the ECS Anywhere installation script in order for exec to be supported on ECS Anywhere. Two changes are required - copying TLS certs and downloading ssm binaries. Both of these will be moved to a directory that will be bind-mounted into the customer's container at start, which is consistent with the current behavior for ECS exec on EC2.

For further context on exec, see:
- [Original ECS Exec proposal](https://github.com/aws/containers-roadmap/issues/1050)
- [AWS Docs on using ECS Exec](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html#ecs-exec-required-iam-permissions)
- [ECS Exec blog post](https://aws.amazon.com/blogs/containers/new-using-amazon-ecs-exec-access-your-containers-fargate-ec2/)

### Implementation details
The changes to installation script are:
1. Move certs. Looking for certs in the [same places as Golang](https://golang.org/src/crypto/x509/root_linux.go) and moving them to the directory that will be bind-mounted
1. Add optional flag for the certs file. If host certs are not in any of searched locations, the customer can provide the path to the certs file 
1. Download SSM binaries. Downloading from new public URL from the SSM team and moving to the directory that will be bind-mounted

### Testing
- Ran script on ubuntu, centos, and debian systems to register to ecs cluster
- Ran tasks on these container instances with exec enabled and used exec
- Passed cert location with the new flag both for an existing and non existing file and verified behavior was as expected

Add exec prerequisites to ecs-anywhere installation script

### Licensing

### DIFF
--- a/scripts/ecs-anywhere-install.sh
+++ b/scripts/ecs-anywhere-install.sh
@@ -35,6 +35,8 @@ usage() {
         (optional) Version of the ECS agent rpm/deb package to use. If not specified, default to latest.
   --skip-registration
         (optional) if this is enabled, SSM agent install and instance registration with SSM is skipped.
+  --certs-file
+        (optional) TLS certs for execute command feature. Defaults to searching for certs in known possible locations.
   --no-start
         (optional) if this flag is provided, SSM agent, docker and ECS agent will not be started by the script despite being installed."
 }
@@ -51,6 +53,7 @@ ECS_VERSION=""
 DEB_URL=""
 RPM_URL=""
 ECS_ENDPOINT=""
+CERTS_FILE=""
 # Whether to check signature for the downloaded amazon-ecs-init package. true unless --skip-gpg-check
 # specified. --skip-gpg-check is mostly for testing purpose (so that we can test a custom build of ecs init package
 # without having to sign it).
@@ -105,6 +108,11 @@ while :; do
     --ecs-endpoint)
         check-option-value "$1" "$2"
         ECS_ENDPOINT="$2"
+        shift 2
+        ;;
+    --certs-file)
+        check-option-value "$1" "$2"
+        CERTS_FILE="$2"
         shift 2
         ;;
     --skip-registration)
@@ -206,12 +214,14 @@ S3_URL_SUFFIX=""
 if grep -q "^cn-" <<< "$REGION"; then
     S3_URL_SUFFIX=".cn"
 fi
+S3_URL="https://s3.${REGION}.amazonaws.com${S3_URL_SUFFIX}"
+SSM_S3_BUCKET="amazon-ssm-$REGION"
 
 if [ -z "$RPM_URL" ]; then
-    RPM_URL="https://s3.${REGION}.amazonaws.com${S3_URL_SUFFIX}/${S3_BUCKET}/$RPM_PKG_NAME"
+    RPM_URL="${S3_URL}/${S3_BUCKET}/$RPM_PKG_NAME"
 fi
 if [ -z "$DEB_URL" ]; then
-    DEB_URL="https://s3.${REGION}.amazonaws.com${S3_URL_SUFFIX}/${S3_BUCKET}/$DEB_PKG_NAME"
+    DEB_URL="${S3_URL}/${S3_BUCKET}/$DEB_PKG_NAME"
 fi
 
 # source /etc/os-release to get the VERSION_ID and ID fields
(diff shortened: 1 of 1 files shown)


scripts/ecs-anywhere-install.sh
```diff
@@ -592,6 +600,75 @@ wait-agent-start() {
...
+install-ssm-binaries-for-exec() {
+    BINARY_VERSION="3.1.338.0"
+    BINARY_PATH="/var/lib/ecs/deps/execute-command/bin/${BINARY_VERSION}"
+    ARCHITECTURE="$(uname -m)"
+
+    echo "Downloading SSM binaries for exec feature"
+    cd && mkdir -p ssm-binaries && cd ssm-binaries
+    case $ARCHITECTURE in
+        'x86_64')
+            curl https://s3.${REGION}.amazonaws.com/amazon-ssm-${REGION}/${BINARY_VERSION}/linux_amd64/amazon-ssm-agent-binaries.tar.gz -o amazon-ssm-agent.tar.gz
```
ubhattacharjya, line 656: For CN partition, I think, this S3 url will be a bit different -> s3.${REGION}.amazonaws.com.cn

scripts/ecs-anywhere-install.sh
```diff
@@ -592,6 +600,75 @@ wait-agent-start() {
...
+        do
+            if  [ -f "$f" ]; then
+                certs=$f
+                break
+            fi
+        done
+    fi
+
+    if [ -z "$certs" ]; then
+        echo "Could not find certificates. Please generate TLS certificates, or provide a valid file path to them"
```
ubhattacharjya, line 636: Can this message ask the user to rerun the script with the --certs-file flag to provide the vaild certificate path?

scripts/ecs-anywhere-install.sh
```diff
@@ -35,6 +35,8 @@ usage() {
         (optional) Version of the ECS agent rpm/deb package to use. If not specified, default to latest.
   --skip-registration
         (optional) if this is enabled, SSM agent install and instance registration with SSM is skipped.
+  --certs-file
+        (optional) file for TLS certs. Defaults to predefined location for a given platform.
```
sharanyad, line 39: Could we add more details such as - TLS certs for Execute command feature

scripts/ecs-anywhere-install.sh
```diff
@@ -592,6 +602,73 @@ wait-agent-start() {
...
+        echo "Using $certs"
+        mkdir -p $CERTS_PATH && cd $CERTS_PATH
+        cp "$certs" tls-ca-bundle.pem
+        chmod 400 tls-ca-bundle.pem
+    fi
+    ok
+}
+
+download-ssm-binaries-exec() {
+    BINARY_VERSION="3.1.338.0"
```
Realmonia, line 654: Is there a specific reason we use this version? Is there a `latest` that we can use?

lydiafilipe: There will be a stable version endpoint available 12/08. We're hardcoding version for now to access what's currently available, but around that date I'll update the script in this branch to use that

Realmonia (approved):

scripts/ecs-anywhere-install.sh
```diff
@@ -592,6 +602,73 @@ wait-agent-start() {
...
+        done
+    fi
+
+    # Copy certs to exec directory
+    if [ -z "$certs" ]; then
+        echo "Could not find certificates. Please rerun with --certs-file and provide a valid path"
+        fail
+    else
+        echo "Using $certs"
+        mkdir -p $CERTS_PATH && cd $CERTS_PATH
```
angelcar, line 646: not sure if `cd` would break other assumptions in the script now or in the future. Perhaps it's more convenient to not change the current dir and use absolute paths?

scripts/ecs-anywhere-install.sh
sparrc: we're currently using 3.1.90.0 in the ecs ami: https://github.com/aws/amazon-ecs-ami/blob/main/variables.pkr.hcl#L59

we should probably use the same version, either by downgrading this version or following up with an ecs ami PR to update the version the AMI uses.

If we do update, might be worth bumping to SSM's latest available version which is 3.1.501.0: https://github.com/aws/amazon-ssm-agent/releases

lydiafilipe: Thanks for pointing that out, it looks like there are a couple places where this might be inconsistent. For now at least I'll update to `3.1.501.0` in this script, though before releasing this we will be updating to a url that just points to the latest version. As part of that, I can look into how we can get that elsewhere and remove some of the hardcoded versions

angelcar (dismissed):

scripts/ecs-anywhere-install.sh
```diff
@@ -592,6 +602,74 @@ wait-agent-start() {
...
+    # Determine certs file
+    certs=""
+    if [ -n "$CERTS_FILE" ] && [ -f "$CERTS_FILE" ]; then
+        certs=$CERTS_FILE
+    else
+        if [ -n "$CERTS_FILE" ]; then
+            echo "Provided certs file does not exist, looking for certs in known possible locations"
+        fi
+        for f in "${CERTS_FILES_LOCATIONS[@]}"
+        do
```
chienhanlin, line 632: Q. Just a thought. Would it provide better UX if we printing out what certs have seen searched?
```
echo "Searching certs file `/etc/ssl/certs/ca-certificates.crt` "
...
echo "Could not find `/etc/ssl/certs/ca-certificates.crt`"
```

lydiafilipe: Interesting point. I can see why some users might want to know these if the script doesn't find the certs. My thinking against doing that is that it would pollute the output a bit to log all of them, and that it's not super useful for the user to know given that these are often already present and easily able to be found just by looking at the script. Also, users can simply specify the location if the certs not found by default.

chienhanlin: I agree doing this will a kind of messing up output, and users can always provide a path if the certs cannot be found.

chienhanlin (approved):

lydiafilipe:
(accidentally removed mkdir $CERTS_PATH in last revision, added that back, which is the only difference)

angelcar (approved):

Realmonia (approved):

PR #462, opened 2021-12-07
add amazon-ecs-volume-plugin to rpm generic package

Realmonia:
### Summary

Add `amazon-ecs-volume-plugin` to `generic-rpm` package

### Implementation details

Mostly the same as https://github.com/aws/amazon-ecs-init/pull/433. Fixed some minor issue, added readme and did some tests.

### Testing

New tests cover the changes:  N/A
* MACIS ecs-a tests passed with .rpm package built by code change
* Install package on a fedora34 instance and successfully run a task with EFS in task definintion.

### Licensing

yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,8 +125,11 @@ BUILDROOT/ecs-agent.tar:
 
 .generic-rpm-done:
 	./scripts/update-version.sh
+	curl -O ${AGENT_URL}
 	cp packaging/generic-rpm/amazon-ecs-init.spec amazon-ecs-init.spec
 	cp packaging/generic-rpm/ecs.service ecs.service
+	cp packaging/generic-rpm/amazon-ecs-volume-plugin.service amazon-ecs-volume-plugin.service
+	cp packaging/generic-rpm/amazon-ecs-volume-plugin.socket amazon-ecs-volume-plugin.socket
 	tar -czf ./sources.tgz ecs-init scripts
 	test -e SOURCES || ln -s . SOURCES
 	rpmbuild --define "%_topdir $(PWD)" -bb amazon-ecs-init.spec
--- a/packaging/generic-rpm/amazon-ecs-init.spec
+++ b/packaging/generic-rpm/amazon-ecs-init.spec
@@ -16,6 +16,7 @@
 %global _cachedir %{_localstatedir}/cache
 %global bundled_agent_version %{version}
 %global no_exec_perm 644
+%global debug_package %{nil}
 
 %ifarch x86_64
 %global agent_image %{SOURCE3}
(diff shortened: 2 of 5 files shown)


packaging/generic-rpm/amazon-ecs-init.spec
```diff
@@ -16,6 +16,8 @@
 %global _cachedir %{_localstatedir}/cache
 %global bundled_agent_version %{version}
 %global no_exec_perm 644
+%global _missing_build_ids_terminate_build 0
```
fenxiong, line 19: for my understanding, what is this used for?

Realmonia: I encountered following error when running `make rpm` on fedora 34.
```
RPM build errors:
    Missing build-id in /home/fedora/go/src/github.com/aws/amazon-ecs-init/BUILDROOT/ecs-init-1.57.0-1.fc33.x86_64/usr/libexec/amazon-ecs-volume-plugin
    Generating build-id links failed
```
This seems to be a new thing on fedora 34. Later realized I was running wrong target but forgot to delete this line. I have tried to remove this line and the build passed. I can remove this from generic-rpm spec. Do we want to add this to rpm spec?

fenxiong: i'm not sure if it's a good idea to let the build pass if the build-id is missing (which seems to be what this macro does). unless you are certain that missing this won't be an issue, i'd suggesting removing it

fenxiong (approved):

Makefile
```diff
@@ -125,8 +125,11 @@ BUILDROOT/ecs-agent.tar:
 
 .generic-rpm-done:
 	./scripts/update-version.sh
+	curl -O ${AGENT_URL}
```
yinyic, line 128: It seems we shouldn't need it https://github.com/aws/amazon-ecs-init/pull/433#discussion_r724363912

packaging/generic-rpm/amazon-ecs-init.spec
```diff
@@ -73,25 +77,32 @@ install -m %{no_exec_perm} %{agent_image} %{buildroot}%{_cachedir}/ecs/
...
 %post
 # Symlink the bundled ECS Agent at loadable path.
 ln -sf %{basename:%{agent_image}} %{_cachedir}/ecs/ecs-agent.tar
 %systemd_post ecs
+%systemd_post amazon-ecs-volume-plugin
 
 %postun
-%systemd_postun
+%systemd_postun ecs.service
+%systemd_postun amazon-ecs-volume-plugin
```
yinyic, line 105: Are we trying to restart the services when upgrade is done? Would `systemd_postun_with_restart` be more suitable? https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd

yinyic: And a more general question - if we already have `post` which gets executed when new package installed, what's the benefit of having `postun`?

Makefile
Realmonia: actually if we don't curl it in install we still need to manually download it, rpm build do not automatically pull resources specified in spec.

packaging/generic-rpm/amazon-ecs-init.spec
Realmonia: Good catch, I take a look at the spec for al2 and we are doing `systemd_postun_with_restart` there.

My thought for question is `postun` get executed after `post` during update so adding it will be safe. I am not sure about the initial intention this is introduced though.

yinyic (approved):

PR #473, opened 2022-01-20
Log what pre-start is doing

sparrc:
### Summary

Logs what the pre-start function in ecs-init is doing, so that us and customers can understand what is taking time during ecs-init package startup. This will log to the ecs service's systemd journal before the agent container starts.

### Testing

amazon-ecs-init binary was manually built and tested.

Enhancement - Add more granular logging to ecs-init prestart process.

### Licensing

### DIFF
--- a/ecs-init/engine/engine.go
+++ b/ecs-init/engine/engine.go
@@ -112,16 +112,19 @@ func (e *Engine) PreStart() error {
 		return err
 	}
 	// Enable use of loopback addresses for local routing purposes
+	log.Info("pre-start: enabling loopback routing")
 	err = e.loopbackRouting.Enable()
 	if err != nil {
 		return engineError("could not enable loopback routing", err)
 	}
 	// Disable ipv6 router advertisements
+	log.Info("pre-start: disabling ipv6 router advertisements")
 	err = e.ipv6RouterAdvertisements.Disable()
 	if err != nil {
 		return engineError("could not disable ipv6 router advertisements", err)
 	}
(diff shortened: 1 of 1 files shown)


ecs-init/engine/engine.go
```diff
@@ -112,16 +112,19 @@ func (e *Engine) PreStart() error {
 		return err
 	}
 	// Enable use of loopback addresses for local routing purposes
+	log.Infof("pre-start: enabling loopback routing")
```
angelcar, line 115: nit: no formatting is required, so `log.Info` would be more appropriate. Not a blocker, though

angelcar (approved):

fenxiong (approved):

angelcar (approved):

ecs-init/engine/engine.go
```diff
@@ -168,6 +176,8 @@ func (e *Engine) PreStartGPU() error {
 	envVariables := docker.LoadEnvVars()
```
yinyic, line 176: Do we want a debug line to print all env vars here? 
We have debug output for user overrides [here](https://github.com/aws/amazon-ecs-init/blob/b2584f1ee3446b767e7ffc37fadf88cdeeec2aee/ecs-init/docker/docker.go#L354), would be good to see the final version too

sparrc: discussed offline, unfortunately ecs-init doesn't currently have the ability to adjust logging level.

PR #474, opened 2022-01-21
build al2 rpm in docker

petderek:
### Summary

This lets us use the command `make rpm-in-docker` to build locally. This
is a good approximation of an authorative build that we can use to test
changes to init.

### Testing

```
make rpm-in-docker
```

### Licensing

### Implementation Details

This is currently hardcoded to al2 -- but we can amend it in the future to include an option to build the generic rpm as well.

### DIFF
--- a/scripts/dockerfiles/build.dockerfile
+++ b/scripts/dockerfiles/build.dockerfile
@@ -1,0 +1,6 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:2
+RUN yum update -y && yum install -y golang make tar rpm-build
+WORKDIR /workspace/amazon-ecs-init
+# we're going to run the build as the non-privileged user, so we need write access to the directory
+RUN chmod -R 777 /workspace
+CMD /bin/bash -c 'make rpm'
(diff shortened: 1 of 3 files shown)


Realmonia (approved):

yinyic (approved):

scripts/dockerfiles/build.dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:2
+RUN yum update -y && yum install -y golang make tar rpm-build
```
angelcar, line 2: I think the version of golang should probably be fixed?

petderek: I like that idea, but since the [spec file](https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs-init.spec#L52) only specifies a minimum, I _think_ that tracking the latest installed version is closer to what actually happens in authoritative builds.

angelcar (approved):

PR #481, opened 2022-02-18
[AL2022] Set GO111MODULE=auto and update ecs-init.spec

chienhanlin:
### Summary
This PR updates `packaging/amazon-linux-ami/ecs-init.spec` and `scripts/gobuild.sh` for packing ecs-init rpm for [Amazon Linux 2022 (AL220)](https://aws.amazon.com/linux/amazon-linux-2022/faqs/) OS.
  
### Implementation details
1. Add `%global debug_package %{nil}` in amazon-linux-ami/ecs-init.spec to fix `RPM build errors: Empty %files file...`.
2. Append an argument `ecs` after %systemd_postun` in amazon-linux-ami/ecs-init.spec to fix `error: The %systemd_postun macro requires some arguments`.
3. Set `GO111MODULE="auto"` in  gobuild.sh to only use go modules if modules exist.
  
### Testing
1. Built ecs-init rpm on an AL2022 instance (x86_64 and aarch64), installed ECS Agent version 1.58 from the rpm, and ran ECS tasks successfully.
2. Built ecs-init rpm in an AL1 and AL2 (x86_64 and aarch64) using Koji, installed ECS Agent version 1.59 from the rpm, and ran ECS tasks successfully.

Support building ecs-init rpm on AL2022 OS.

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ecs-init.spec
+++ b/packaging/amazon-linux-ami/ecs-init.spec
@@ -23,6 +23,7 @@
 %global _cachedir %{_localstatedir}/cache
 %global bundled_agent_version %{version}
 %global no_exec_perm 644
+%global debug_package %{nil}
 
 %ifarch x86_64
 %global agent_image %{SOURCE3}
@@ -224,7 +225,7 @@ ln -sf %{basename:%{agent_image}} %{_cachedir}/ecs/ecs-agent.tar
 %systemd_post amazon-ecs-volume-plugin.service
 
 %postun
-%systemd_postun
+%systemd_postun ecs
 %systemd_postun_with_restart amazon-ecs-volume-plugin
 
 %else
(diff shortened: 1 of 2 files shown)


packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -23,6 +23,7 @@
 %global _cachedir %{_localstatedir}/cache
 %global bundled_agent_version %{version}
 %global no_exec_perm 644
+%global debug_package %{nil}
```
Realmonia, line 26: What is the disabling intended for?

Realmonia, line 26: Q: Is there anything we lost in build logs after adding this?

Realmonia (approved):

packaging/amazon-linux-ami/ecs-init.spec
chienhanlin: Thanks @yinyic for adding the Debuginfo! 
When we build rpms with internal Koji, the building target platforms is set to `noarch` by default; therefore, debuginfo package is not created according to this [section](https://docs.fedoraproject.org/en-US/packaging-guidelines/Debuginfo/#_missing_debuginfo_packages).

> **Missing debuginfo packages**
It is normal for noarch package builds to not produce a debuginfo package. If it’s missing in other cases (where it has not been explicitly disabled), something’s wrong. One such case is a [missing %build section](https://bugzilla.redhat.com/192422) with some rpmbuild versions.

yinyic (approved):

PR #498, opened 2022-03-31
check ipv4 routes for default network interface, only fall back to eth0 if none can be found

yinyic:
### Summary
Currently we default to `eth0` when creating iptables rules for blocking offhost introspection endpoint access. 

However, with the introduction of predictable network interface names, primary network interfaces may no longer have the name `eth0`. This has been observed on a few EC2 Nitro instances when launched with AL2022 AMI (systemd > v197) where the primary network interfaces were named such as `ens3`, `ens5`, etc. 

Reference this [page](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) for more information on predictable network interface names. Quoting the summary -
> Starting with v197 systemd/udev will automatically assign predictable, stable network interface names for all local Ethernet, WLAN and WWAN interfaces. This is a departure from the traditional interface naming scheme ("eth0", "eth1", "wlan0", ...), but should fix real problems. 

This PR attempts to remedy this by identifying the default network interface name through route file. If none could be found, we fall back to `eth0` to be backward-compatible (in the rare case that it might affect some customer with a special routing setup that's previously working).

### Implementation details

`/proc/net/route` file stores IPv4 kernel routing table, where each row (except header) represent a route, e.g. 
```
$ sudo cat /proc/net/route 
Iface	Destination	Gateway 	Flags	RefCnt	Use	Metric	Mask		MTU	Window	IRTT                                                       
ens5	00000000	01201FAC	0003	0	0	512	00000000	0	0	0                                                                             
ens5	00201FAC	00000000	0001	0	0	0	00F0FFFF	0	0	0                                                                               
ens5	01201FAC	00000000	0005	0	0	512	FFFFFFFF	0	0	0   
```
We will be looking for the interface that's mapped to the all-zero destination (the default route).

This is equivalent to using `ip route` command
```
$ ip route show to default
default via 172.31.32.1 dev ens5 proto dhcp src 172.31.32.21 metric 512
```
However introducing a new utility command seems a bit overkill if we are only using it once. More importantly, I found `ip route` sometimes not showing full routes in the case of IPv6 (`ip route -6`) which can potentially cause gaps if we are to add IPv6 support.

### Testing

New tests cover the changes:  new unit tests

[Enhancement] check ipv4 routes for default network interface instead of defaulting to eth0

### Licensing


ecs-init/exec/iptables/iptables.go
```diff
@@ -189,18 +203,52 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
+// Iface   Destination Gateway     Flags   RefCnt  Use Metric  Mask   MTU Window  IRTT
+// ens5    00000000    01201FAC    0003    0   		0   512 00000000    0   0   	0
+// ...
+//
+// 1st column contains interface name
+// 2nd column contains destination network in hex
+var getDefaultNetworkInterfaceIPv4 = func() (string, error) {
+	input, err := os.Open(ipv4RouteFile)
+	if err != nil {
+		return "", fmt.Errorf("Could not get IPv4 route input: %v", err)
```
fenxiong, line 232: nit: error message should start with lower case; ditto for line 251 and line 83

ecs-init/exec/iptables/iptables.go
```diff
@@ -189,18 +203,52 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
+	}
+	defer input.Close()
+	return scanIPv4RoutesForDefaultInterface(input)
+}
+
+func scanIPv4RoutesForDefaultInterface(input io.Reader) (string, error) {
+	scanner := bufio.NewScanner(input)
+	for scanner.Scan() {
+		line := scanner.Text()
+		fmt.Println(line)
```
fenxiong, line 242: should this line be removed?

ecs-init/exec/iptables/iptables.go
```diff
@@ -189,18 +203,52 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
 		"-p", "tcp",
-		"-i", getOffhostIntrospectionInterface(),
+		"-i", defaultOffhostIntrospectionInterface,
 		"--dport", agentIntrospectionServerPort,
 		"-j", "DROP",
 	}
 }
 
-func getOffhostIntrospectionInterface() string {
+func getOffhostIntrospectionInterface() (string, error) {
```
fenxiong, line 212: i think the documentation for ECS_OFFHOST_INTROSPECTION_INTERFACE_NAME env in README should be updated as part of this change

fenxiong (commented):
lgtm minor comments

ecs-init/exec/iptables/iptables.go
yinyic: Yes! Sorry forgot to take it out :)

ecs-init/exec/iptables/iptables.go
yinyic: You are right, which reminds me we will also need to update the public Agent configuration [doc](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) when the change is out.

fenxiong (approved):

ecs-init/exec/iptables/iptables.go
```diff
@@ -69,6 +78,11 @@ func NewNetfilterRoute(cmdExec exec.Exec) (*NetfilterRoute, error) {
 		return nil, err
 	}
 
+	defaultOffhostIntrospectionInterface, err = getOffhostIntrospectionInterface()
```
sparrc, line 81: I'm wondering if we should have a fallback here for if this fails? Do we know for sure that we'll always have access to `/proc/net/route` and it will always look as expected here?

ecs-init/exec/iptables/iptables.go
```diff
@@ -189,18 +203,51 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
+//
+// 1st column contains interface name
+// 2nd column contains destination network in hex
+var getDefaultNetworkInterfaceIPv4 = func() (string, error) {
+	input, err := os.Open(ipv4RouteFile)
+	if err != nil {
+		return "", fmt.Errorf("could not get IPv4 route input: %v", err)
+	}
+	defer input.Close()
+	return scanIPv4RoutesForDefaultInterface(input)
```
sparrc, line 235: Will this fail if ipv4 is disabled?

ecs-init/exec/iptables/iptables.go
yinyic: This file is created by kernel at boot time as part of procfs so the content should be kernel's knowledge of IPv4 routes. I'm not sure if there should be a fallback or what a reasonable fallback would be if we couldn't find a default route there. 

I checked latest AL1, AL2, AL2022, Ubuntu 20.04, SUSE SLES 15, and Debian 10, all have the file present, readable by everyone (permission 444), and contain the default route. Any suggestion on other distros I should check?

ecs-init/exec/iptables/iptables.go
yinyic: I suppose user can explicitly delete the default IPv4 route (or maybe there's a way to tell kernel not to populate the route table on boot?). In such case it's not so meaningful setting up the drop rule anyways (as it's targeting IPv4 only). Maybe we can just log an error and move on? What do you think?

ecs-init/exec/iptables/iptables.go
sparrc: That seems like a reasonable range of distros to test. I tried to find some other distro with a different file but it does seem fairly consistent

ecs-init/exec/iptables/iptables.go
sparrc: That's true I didn't realize this was only for the offhost introspection drop rule anyways. I guess I only wonder that if it fails should we fall back to the previous behavior (harcoded eth0) in the rare case that it might affect some customer with a special routing setup that's currently working.

yinyic: Yeah okay I guess I don't see any harms doing that. 

Creating a drop rule with `-i eth0` on an instance without eth0 device actually works, although the rule probably never will be picked up.

sparrc (approved):

